PR #2, opened 2017-04-07
WIP postgreSQL compatibile

davecramer:


ivannovick:
nice to see a new repo with activity, i am happy to help review if you don't get other people @davecramer

Eulerizeit:
@ictmalili Are we planning to officially support Postgres? I personally think we should. This means moving on this PR and adding Postgres to the CI. 

The advantage of this would be that this could be a real community project drawing developers from outside Pivotal. Otherwise this will only be developed by Pivotal and can only go as far as the developers allocated to the project can take it alone.

schubert:
+1 I'd strongly support making sure plcontainer works with postgres as well like all other pl components we want wider adoption. cc @ivannovick

ictmalili:
@davecramer @Eulerizeit Which version of Postgres is this PR for?

davecramer:
@ictmalili it should work with any recent version. I think I tried 9.6

Eulerizeit:
@ictmalili is there a particular issue with this PR that can be addressed so it can be merged?

src/common/comm_channel.c
```diff
@@ -331,8 +331,10 @@ static int send_udt(plcConn *conn, plcType *type, plcUDT *udt) {
 }
 
 static int receive_message_type(plcConn *conn, char *c) {
+
+	int res = plcBufferReceive(conn, 1);
```
paul-guo-: I noticed almost of the previous code use space for indent, but I think we might want the code in the future use tab for indent to follow the pg convention. So I'm fine with the tab indent.

src/message_fns.c
```diff
@@ -33,7 +33,13 @@ interpreted as representing official policies, either expressed or implied, of t
 /* Greenplum headers */
 #include "postgres.h"
 #include "executor/spi.h"
+#include "access/htup_details.h"
 #include "access/transam.h"
+#include "catalog/pg_proc.h"
+#include "catalog/pg_type.h"
+#include "utils/array.h"
+#include "utils/syscache.h"
+#include "utils/builtins.h"
```
paul-guo-: Why do we add these new header files? They are required for compiling on pg?

src/plc_configuration.c
```diff
@@ -254,6 +255,8 @@ int plc_read_container_config(bool verbose) {
      */
     LIBXML_TEST_VERSION
 
+	data_directory = GetConfigOption("data_directory", false, true);
```
paul-guo-: This violates the current GetConfigOption() definition in gpdb, i.e.

const char *
GetConfigOption(const char *name)

We should have different calls for pg and gp.

tests/Makefile
```diff
@@ -9,12 +9,12 @@
 PGXS := $(shell pg_config --pgxs)
 include $(PGXS)
 
-REGRESS_OPTS = --dbname=$(PL_TESTDB) --init-file=./init_file
+GPDBREGRESS_OPTS = --dbname=$(PL_TESTDB) --init-file=./init_file
+PSQLREGRES_OPTS = --dbname=$(PL_TESTDB)
```
paul-guo-: GPDBREGRESS_OPTS -> GPDB_REGRESS_OPTS (Later seems to be better).
PSQLREGRES_OPTS -> PSQL_REGRES_OPTS or why not PG_REGRES_OPTS?

tests/Makefile
```diff
@@ -71,12 +70,12 @@ prepare:
 .PHONY: tests
 tests: prepare
 	$(top_builddir)/src/test/regress/pg_regress \
-				--psqldir=$(PSQLDIR) $(REGRESS_OPTS) $(REGRESS_GPDB5)
+				--psqldir=$(PSQLDIR) $(GPDBREGRESS_OPTS) $(REGRESS_GPDB5)
 
-.PHONY: tests4
-tests4: prepare
+.PHONY: postgres
```
paul-guo-: postgres -> tests_pg? The "postgres" target name is misleading.

PR #5, opened 2017-08-01
Change submodule of plcontainer to use https instead of ssh.

zhangh43:
submodule should use https instead of ssh. Or user cannot read only access submodule.


paul-guo-:
+1

paul-guo-:
This benefits auto-pipeline work.

PR #6, opened 2017-08-15
Fix compile error "../common/comm_messages.c:254: error: comparison o…

paul-guo-:
…f unsigned expression >= 0 is always true" when using gcc 4.4.7 in my environment.

### DIFF
--- a/src/common/comm_messages.c
+++ b/src/common/comm_messages.c
@@ -243,13 +243,21 @@ int plc_get_type_length(plcDatatype dt) {
     return res;
 }
 
+static const char *plcDatatypeName[] =
+{
+                        "PLC_DATA_INT1",
+                        "PLC_DATA_INT2",
+                        "PLC_DATA_INT4",
+                        "PLC_DATA_INT8",
+                        "PLC_DATA_FLOAT4",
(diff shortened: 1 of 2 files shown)


src/common/comm_messages.c
```diff
@@ -243,13 +243,19 @@ int plc_get_type_length(plcDatatype dt) {
...
+                        "PLC_DATA_INT2",
+                        "PLC_DATA_INT4",
+                        "PLC_DATA_INT8",
+                        "PLC_DATA_FLOAT4",
+                        "PLC_DATA_FLOAT8",
+                        "PLC_DATA_TEXT",
+                        "PLC_DATA_ARRAY",
+                        "PLC_DATA_UDT",
+                        "PLC_DATA_BYTEA",
+                        "PLC_DATA_INVALID"};
```
stanlyxiang: Better to have "};" in new line  ?

zhangh43:
+1

src/common/comm_messages.c
paul-guo-: Agree. Also reformat it more.

paul-guo-:
Merged.

PR #9, opened 2017-08-18
Makefile Refactor and gppkg build script

HaozhouWang:
1. refactored the Makefile, remove unnecessary make targets. 
2. The gppkg script also refactored and now support OSS gpdb.
3. Add complier flag to adjust docker API version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,9 +53,9 @@ PLCONTAINERDIR = $(DESTDIR)$(datadir)/plcontainer
 
 override CFLAGS += -Werror
 
-# detected the docker API version
-OS_VERSION=$(shell cat /etc/redhat-release | sed s/.*release\ // | sed s/\ .*// | awk -F '.' '{print $$1}' )
-ifneq ($(OS_VERSION), 7)
+# detected the docker API version, only for centos 6
+RHEL_MAJOR_OS=$(shell cat /etc/redhat-release | sed s/.*release\ // | sed s/\ .*// | awk -F '.' '{print $$1}' )
+ifeq ($(RHEL_MAJOR_OS), 6)
   override CFLAGS +=  -DDOCKER_API_LOW
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -9,18 +9,13 @@ MODULE_big = plcontainer
 # Release versions
 include release.mk
 
-ifeq ($(CIBUILD),1)
-  IMAGE_TAG=devel
-else
-  IMAGE_TAG=$(PLCONTAINER_VERSION).$(PLCONTAINER_RELEASE)-$(PLCONTAINER_IMAGE_VERSION)
-endif
-
 # Directories
 SRCDIR = ./src
 MGMTDIR = ./management
-DOCKERFILEDIR = ./dockerfiles
 PYCLIENTDIR = src/pyclient/bin
 RCLIENTDIR = src/rclient/bin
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -56,9 +51,17 @@ endif
 
 PLCONTAINERDIR = $(DESTDIR)$(datadir)/plcontainer
 
+override CFLAGS += -Werror
+
+# detected the docker API version
+OS_VERSION=$(shell cat /etc/redhat-release | sed s/.*release\ // | sed s/\ .*// | awk -F '.' '{print $$1}' )
+ifneq ($(OS_VERSION), 7)
+  override CFLAGS +=  -DDOCKER_API_LOW
+endif
```
paul-guo-: I'd suggest warn if it is not redhat distribution.

package/Makefile
```diff
@@ -5,14 +5,14 @@ include $(PGXS)
 
 include ../release.mk
 
-OS=$(word 1,$(subst _, ,$(BLD_ARCH)))
+OS_PL=$(shell cat /etc/redhat-release | sed s/.*release\ // | sed s/\ .*// | awk -F '.' '{print $$1}' )
```
paul-guo-: Maybe MAJOR_OS is better?

src/plc_docker_curl_api.c
```diff
@@ -24,8 +24,11 @@
 static char *plc_docker_socket = "/var/run/docker.sock";
 
 // URL prefix specifies Docker API version
-static char *plc_docker_url_prefix = "http:/v1.21";
-
+#ifdef DOCKER_API_LOW
+    static char *plc_docker_url_prefix = "http:/v1.19";
+#else
+    static char *plc_docker_url_prefix = "http:/v1.21";
+#endif
```
paul-guo-: We need to modify for the non-curl version also.

Eulerizeit:
@ivannovick do we need to worry about Ubuntu or SuSE 12 in the near future?

davecramer:
@Eulerizeit why limit it at all ? it works fine as is

Makefile
HaozhouWang: Thanks for your review, changed to only for centos 6 docker version

package/Makefile
HaozhouWang: Thanks for your review, changed to MAJOR_OS

src/plc_docker_curl_api.c
HaozhouWang: Oh sorry, this is my fault. I will update the non-curl version as well.

HaozhouWang:
We changed the Makefile to use lower version docker API for centos 6. If we need support Ubuntu or SuSE12, we can open a new PR to update the Makefile to detect the OS version and docker API version in that OS.

HaozhouWang:
@paul-guo- Yes, exactly.
We currently build the plcontainer.so firstly, and then clean the  src/common/*.o, and build the clients.

stanlyxiang:
question: do we need to update license header in our pr instead of a separate pr to update all  ? 
@HaozhouWang @paul-guo-

paul-guo-:
Agree. In the long run, we could generate docker version constants for various OSes. If we have more and more dynamic variables, we could consider add configure.

HaozhouWang:
@stanlyxiang +1 to update in our pr

PR #10, opened 2017-08-18
Set up framework plcontainer pipeline to run funtion test and performance test.

zhangh43:
This PR setup the plcontainer pipeline to build plcontainer, build client docker image, run function test and performance test.


dockerfiles/python/centos7/ds/install_python.sh
```diff
@@ -0,0 +1,7 @@
+pushd /opt/ds
+rpm2cpio DataSciencePython-*.x86_64.rpm| cpio -div
+mv temp/ext/DataSciencePython/* ./
```
paul-guo-: You could install using rpm command directly.

concourse/scripts/plcontainer_clients_build.sh
```diff
@@ -22,35 +22,20 @@ popd
...
 make clean && make
 popd
-cp src/pyclient/client    ../plcontainer_clients_build/python34/client
-unset PYTHON
-export LD_LIBRARY_PATH=/usr/local/anaconda/lib
-export PYTHONHOME=/usr/local/anaconda
-export PATH=/usr/local/anaconda/bin:$PATH
-pushd src/pyclient
-make clean && make
+cp src/pyclient/client     ../plcontainer_clients_build/python26/client
```
paul-guo-: Why python26?

concourse/scripts/perf_py1.sql
```diff
@@ -0,0 +1,4 @@
+\timing
+explain analyze select pylog100_py() from a;
+explain analyze select pylog100_py() from a4;
+explain analyze select pylog100_py() from a5;
```
paul-guo-: I think we do not need "explian analyze".

concourse/scripts/plcontainer_clients_build.sh
davecramer: probably because it was originally written to support gpdb4 so we can get rid of python26

concourse/scripts/plcontainer_gpdb_build.sh
```diff
@@ -5,12 +5,12 @@ set -x
 GPDBBIN=$1
 OUTPUT=$2
 MODE=$3
-
-cp $GPDBBIN/$GPDBBIN.tar.gz /usr/local
-pushd /usr/local
-tar zxvf $GPDBBIN.tar.gz
+mkdir /usr/local/greenplum-db-devel
```
stanlyxiang: sometimes the path is /usr/local/greenplum-db while sometimes the path is /usr/local/greenplum-db-devel. Is there any reason to use 2 different path ? though there are in separated files.  
If no, can we unify them ?

src/common/comm_messages.c
```diff
@@ -259,5 +259,12 @@ static const char *plcDatatypeName[] =
...
-    return ((unsigned int) dt <= PLC_DATA_INVALID) ? plcDatatypeName[dt] : "UNKNOWN";
+    const char * types[] = {"PLC_DATA_INT1", "PLC_DATA_INT2", "PLC_DATA_INT4", "PLC_DATA_INT8",
+                            "PLC_DATA_FLOAT4", "PLC_DATA_FLOAT8",
+                            "PLC_DATA_TEXT",
+                            "PLC_DATA_ARRAY",
+                            "PLC_DATA_UDT",
+                            "PLC_DATA_BYTEA",
+                            "PLC_DATA_INVALID"};
+    return (dt <= 10) ? types[dt] : "UNKNOWN";
 }
```
zhangh43: merge error, fixed it

concourse/scripts/plcontainer_clients_build.sh
zhangh43: I remove this file, since it is no longer used by the pipeline.

concourse/scripts/perf_py1.sql
zhangh43: better to use explain analyze in performance test. since we could get more hints from the output.

concourse/scripts/plcontainer_gpdb_build.sh
zhangh43: Some legacy test use path /usr/local/greenplum-db. I have deleted all of them to make the pipeline clear.

zhangh43:
@paul-guo- I think we should follow gpdb repo not to add file header for concourse related files.

concourse/scripts/perf_py1.sql
paul-guo-: I'm not sure about the overhead of "explain analyze" vs direct run.

If the test time is not long, you could run it 2nd time without "explain analyze".

By the way, I'm a bit worried about how we evaluate the final performance since it is very possible that all the data is in fast RAM before testing, but let's have the framework at first.

concourse/scripts/plcontainer_gpdb_build.sh
paul-guo-: Maybe we should test with a more real usual path (i.e. without -devel) to avoid potential issues in the future? How do you think?

zhangh43: It's just the path where the gpdb binary is located. Code should never have any dependency on this path( in fact we should always read it from env variable GP_HOME)

concourse/scripts/perf_py1.sql
zhangh43: The overhead is small. 
Data in RAM is correct. We could run it many times and calculate mean without the first run. To clear RAM cache every time is another way.

paul-guo-:
+1

HaozhouWang:
+1

PR #11, opened 2017-08-21
Use exit() instead of abort() in client so that atxit() could be used.

paul-guo-:
We will want some callbacks registered in atexit().

By now, we do not use fine-grained level handling code in gp/pg;
we just simple call exit() for level >= ERROR. We could allow
abort() if needed in the future.


HaozhouWang:
LGTM

zhangh43:
+1

PR #12, opened 2017-08-21
Enable -Wextra flag in plcontainer

HaozhouWang:
1. Enable -Wextra flag and fix related errors.
2. Using pg_attribute_unused() instead of UNUSED.


paul-guo-:
+1

PR #13, opened 2017-08-21
Fix make failed after make clients due to common objects not cleaned.

stanlyxiang:
Add clean-clients target to enable clean clients object files when 'make clean'.

Update rclient to latest commit.


HaozhouWang:
+1

paul-guo-:
+1

PR #15, opened 2017-08-22
Fix container name problem and rename struct to avoid ambiguousness

stanlyxiang:
1.replace magic number
2.check container name instead of confused container name rule
3.rename plcContainer -> plcContainerConf to avoid ambiguousness

### DIFF
--- a/src/containers.c
+++ b/src/containers.c
@@ -1,7 +1,6 @@
 /*------------------------------------------------------------------------------
  *
- *
- * Copyright (c) 2016, Pivotal.
+ * Copyright (c) 2017-Present Pivotal Software, Inc
  *
  *------------------------------------------------------------------------------
  */
@@ -147,7 +146,7 @@ plcConn *find_container(const char *image) {
     return NULL;
 }
 
-plcConn *start_container(plcContainerConf *cont) {
+plcConn *start_container(plcContainerConf *conf) {
     int port;
     unsigned int sleepus = 25000;
     unsigned int sleepms = 0;
@@ -170,7 +169,7 @@ plcConn *start_container(plcContainerConf *cont) {
         return conn;
     }
 
-    res = plc_docker_create_container(sockfd, cont, &dockerid);
+    res = plc_docker_create_container(sockfd, conf, &dockerid);
     if (res < 0) {
         elog(ERROR, "Cannot create Docker container");
         return conn;
@@ -233,7 +232,7 @@ plcConn *start_container(plcContainerConf *cont) {
                     CONTAINER_CONNECT_TIMEOUT_MS);
         conn = NULL;
     } else {
-        insert_container(cont->name, dockerid, conn);
+        insert_container(conf->name, dockerid, conn);
     }
 
     pfree(dockerid);
@@ -283,7 +282,6 @@ static inline bool is_whitespace (const char c) {
 char *parse_container_meta(const char *source) {
     int first, last, len;
     char *name = NULL;
-    int nameptr = 0;
 
     first = 0;
     len = strlen(source);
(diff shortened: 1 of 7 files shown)


src/containers.c
```diff
@@ -345,5 +351,24 @@ char *parse_container_meta(const char *source) {
...
     name[nameptr] = '\0';
 
+    int regt = check_container_name(name);
+    if (regt == -1) {
+        lprintf(ERROR, "Container name '%s' contains illegal character for container.", name);
+    }
+
     return name;
 }
+
```
HaozhouWang: It could be better to add some comments about this function

src/containers.c
```diff
@@ -345,5 +351,24 @@ char *parse_container_meta(const char *source) {
     }
     name[nameptr] = '\0';
 
+    int regt = check_container_name(name);
+    if (regt == -1) {
```
HaozhouWang: maybe using boolean instead of -1

src/containers.c
```diff
@@ -320,20 +321,25 @@ char *parse_container_meta(const char *source) {
...
-    name = (char*)pmalloc(last-first+1);
-    while (first < last && source[first] != ':') {
-        if (!is_whitespace(source[first])) {
-            name[nameptr] = source[first];
-            nameptr++;
-        }
+    name = (char*)pmalloc(last-first+1 + 1);
+    while (first <= last ){
+        name[nameptr] = source[first];
+        nameptr++;
```
paul-guo-: Use memcpy instead.

src/containers.c
```diff
@@ -320,20 +321,25 @@ char *parse_container_meta(const char *source) {
...
+     * the character length of name is last-first+1
+     * +1 for terminator
      */
-    name = (char*)pmalloc(last-first+1);
-    while (first < last && source[first] != ':') {
-        if (!is_whitespace(source[first])) {
-            name[nameptr] = source[first];
-            nameptr++;
-        }
+    name = (char*)pmalloc(last-first+1 + 1);
```
paul-guo-: last - first + 1 + 1

src/containers.c
```diff
@@ -345,5 +351,24 @@ char *parse_container_meta(const char *source) {
...
+    regex_t    re;
+    if (regcomp(&re, "^[a-zA-Z0-9][a-zA-Z0-9_.-]*$", REG_EXTENDED | REG_NOSUB | REG_NEWLINE) != 0) {
+        return -1;
+    }
+    status = regexec(&re, name, (size_t) 0, NULL, 0);
+    regfree(&re);
+    if (status != 0) {
+        return -1;
+    }
+    return 1;
```
paul-guo-: I'd it returns 0 for OK case which is the convention of UNIX.

src/plcontainer.c
```diff
@@ -170,7 +170,7 @@ static plcProcResult *plcontainer_get_result(FunctionCallInfo  fcinfo,
     name = parse_container_meta(req->proc.src);
     conn = find_container(name);
     if (conn == NULL) {
-        plcContainer *cont = NULL;
+        plcContainerConf *cont = NULL;
```
paul-guo-: Better cont -> conf for all?

src/containers.c
stanlyxiang: just follow other functions.  -1 for failure and 0 for success.

paul-guo-:
+1

davecramer:
What issue is this addressing ?

stanlyxiang:
@davecramer 
before this fix: 
I can define container name in function like this:   #container: con  tainer _  name.  Plcontainer will find container definition in xml file use name "container_name"( remove blank spaces and combine them together. However I must define "container_name" (without spaces) in xml files. This mismatch is  too confused.  
after this fix: container name in function definition and xml file should be same.  The name rule follow the docker container name convention. 

Do you have idea about the name convention ? Is it too strict ?

stanlyxiang:
@davecramer 
Thank you Dave. We will explain what PRs do in the PRs so that everyone can know the problems and fixes.

PR #17, opened 2017-08-24
Fix multiple shared dirs not work due to lack comma.

stanlyxiang:
Fix multiple shared dirs not work due to lack comma.
Fix docker api plcContainer struct missed rename work.


src/plc_configuration.c
```diff
@@ -357,16 +357,20 @@ char *get_sharing_options(plcContainerConf *conf) {
...
         volumes = palloc(conf->nSharedDirs * sizeof(char*));
         for (i = 0; i < conf->nSharedDirs; i++) {
             volumes[i] = palloc(10 + strlen(conf->sharedDirs[i].host) +
                                  strlen(conf->sharedDirs[i].container));
+            if (i > 0 && comma == ' '){
+                comma = ',';
+            }
             if (conf->sharedDirs[i].mode == PLC_ACCESS_READONLY) {
-                sprintf(volumes[i], "\"%s:%s:ro\"", conf->sharedDirs[i].host,
+                sprintf(volumes[i], " %c\"%s:%s:ro\"", comma, conf->sharedDirs[i].host,
```
HaozhouWang: I think it could be better to use the string in "access" in xml instead of hard coded 'ro' and 'rw' here. As we may need to support  another access type such as 'Z'.

src/plc_configuration.c
```diff
@@ -357,16 +357,20 @@ char *get_sharing_options(plcContainerConf *conf) {
         int totallen = 0;
         char *pos;
         int i;
+        char comma = ' ';
 
         volumes = palloc(conf->nSharedDirs * sizeof(char*));
         for (i = 0; i < conf->nSharedDirs; i++) {
             volumes[i] = palloc(10 + strlen(conf->sharedDirs[i].host) +
                                  strlen(conf->sharedDirs[i].container));
+            if (i > 0 && comma == ' '){
```
paul-guo-: No need to check comma == ' '?

src/plc_configuration.c
stanlyxiang: Agree. But we can do it in next PR because I think there are a lot hardcode problem, not only "ro" and "rw". We could fix them together.
This PR fix the multiple shared dirs problems and the struct rename problem which all result in bugs or compile failure.

src/plc_configuration.c
stanlyxiang: removed

paul-guo-:
+1

PR #19, opened 2017-08-28
Initial unix domain socket support

paul-guo-:
I will add test cases if needed.


src/common/comm_server.c
```diff
@@ -51,6 +54,62 @@ int start_listener() {
 }
 
 /*
+ * Functoin binds the socket and starts listening on it: unix domain socket.
```
davecramer: s/b Function

src/common/comm_server.c
```diff
@@ -51,6 +54,62 @@ int start_listener() {
...
+#endif
+    if (listen(sock, 10) == -1) {
+        lprintf(ERROR, "Cannot listen the socket: %s", strerror(errno));
+    }
+
+    return sock;
+}
+
+/*
  * Fuction waits for the socket to accept connection for finite amount of time
```
davecramer: Different spelling error for Fuction

src/plc_configuration.c
```diff
@@ -378,20 +392,39 @@ char *get_sharing_options(plcContainerConf *conf) {
             totallen += strlen(volumes[i]);
         }
 
-        res = palloc(totallen + 2 * conf->nSharedDirs);
+		if (!conf->isNetworkConnection) {
```
davecramer: indention ??

paul-guo-:
@davecramer typo were fixed. Thanks.

For indentation, I think we should follow the pg convention which seems to be using tab (with 4 space wideth) for indentation. Later we might replace all other space indentation so during development we will use tab for new code change.

src/common/comm_connectivity.c
```diff
@@ -370,14 +376,67 @@ plcConn *plcConnect(int port) {
...
+    tv.tv_sec  = 0;
+    tv.tv_usec = 500000;
+    setsockopt(sock, SOL_SOCKET, SO_RCVTIMEO, (char *)&tv, sizeof(struct timeval));
+
+    result = plcConnInit(sock);
+
+    return result;
+}
+
+
```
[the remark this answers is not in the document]
paul-guo-: Done. Thanks.

paul-guo-:
@stanlyxiang 

I changed the code so there is no need to fix R code. Also memset the conf struct so use_network has an explicit value. Fixed another small bug additionally.

Thanks for reviewing.

paul-guo-:
I'll merge them at first. Currently pg_regress runs with this setting by default. There will have some extra security related things need to do next.

PR #20, opened 2017-08-29
plcontainer can start any function execution env with extensible framework

stanlyxiang:
…ework.


src/plc_backend_api.c
```diff
@@ -0,0 +1,64 @@
...
+    switch (imptype) {
+        case DOCKER_CONTAINER:
+            plc_docker_init(&CurrentPLCImp);
+            break;
+        default:
+            Assert(false);
+    }
+}
+
+int plc_connect(void){
```
paul-guo-: I'm wondering whether we should use rename to plc_backend_connect or other names since we have plcDisconnect() kinda functions which make they are  a bit confusing. So as to other function names.

src/plc_backend_api.c
```diff
@@ -0,0 +1,64 @@
...
+    CurrentPLCImp.create = NULL;
+    CurrentPLCImp.start = NULL;
+    CurrentPLCImp.kill = NULL;
+    CurrentPLCImp.inspect = NULL;
+    CurrentPLCImp.wait = NULL;
+    CurrentPLCImp.delete_backend = NULL;
+    CurrentPLCImp.disconnect = NULL;
+
+    switch (imptype) {
+        case DOCKER_CONTAINER:
```
paul-guo-: So DOCKER_CONTAINER -> BACKEND_DOCKER? See below comment.

src/plc_docker_common.c
```diff
@@ -0,0 +1,25 @@
+/*------------------------------------------------------------------------------
+ *
+ * Copyright (c) 2016-Present Pivotal Software, Inc
```
paul-guo-: 2016->2017

src/plc_backend_api.c
```diff
@@ -0,0 +1,64 @@
...
+            plc_docker_init(&CurrentPLCImp);
+            break;
+        default:
```
paul-guo-: Maybe we should just elog(ERROR) to tell that the type has not been implemented yet.

src/plc_backend_api.c
```diff
@@ -0,0 +1,64 @@
...
+ * Copyright (c) 2016-Present Pivotal Software, Inc
+ *
+ *------------------------------------------------------------------------------
+ */
+
+#include "plc_backend_api.h"
+
+#include "plc_docker_common.h"
+
+PLC_FunctionEntriesData CurrentPLCImp;
```
paul-guo-: It could be static.

src/plc_backend_api.c
```diff
@@ -0,0 +1,64 @@
...
+void plc_prepareImplementation(enum PLC_BACKEND_TYPE imptype) {
+    /* Initialize resource broker implement handlers. */
+    CurrentPLCImp.connect = NULL;
+    CurrentPLCImp.create = NULL;
+    CurrentPLCImp.start = NULL;
+    CurrentPLCImp.kill = NULL;
+    CurrentPLCImp.inspect = NULL;
+    CurrentPLCImp.wait = NULL;
+    CurrentPLCImp.delete_backend = NULL;
+    CurrentPLCImp.disconnect = NULL;
```
paul-guo-: Use memset instead?

src/plc_backend_api.h
```diff
@@ -0,0 +1,57 @@
...
+    PLC_FPTR_delete      delete_backend;
+    PLC_FPTR_disconnect  disconnect;
+};
+
+typedef struct PLC_FunctionEntriesData  PLC_FunctionEntriesData;
+typedef struct PLC_FunctionEntriesData *PLC_FunctionEntries;
+
+enum PLC_BACKEND_TYPE {
+    DOCKER_CONTAINER = 0,
+    UNIMPLEMENT_TYPE
```
paul-guo-: Maybe we should add other types here although we do not support them at this moment?
BACKEND_GARDEN, BACKEND_PROCESS?

src/containers.c
```diff
@@ -76,10 +72,10 @@ static void cleanup(char *dockerid, char *uds_fn) {
             /* Connect to the Docker API and execute "wait" command for the
```
paul-guo-: Better replace all of the docker stuffs in this function.

src/plc_backend_api.c
stanlyxiang: wanna to be more clear with followup interface functions. also consistent with plc_$type_init  function since some backend type only have part of these interfaces.

src/plcontainer.c
```diff
@@ -1,6 +1,6 @@
 /*------------------------------------------------------------------------------
  *
- * Copyright (c) 2017-Present Pivotal Software, Inc
+ * Copyright (c) 2016-Present Pivotal Software, Inc
```
zhangh43: why 2016?

stanlyxiang: this old file was added at 2016 and I updated to wrong 2017 in previous pr, so I reverted to 2016 by the way.

stanlyxiang: old files are 2016 and new added files are 2017.

zhangh43: make sense

paul-guo-:
+1

PR #22, opened 2017-08-30
Add gcc -g for pyclient.

paul-guo-:
This benefits debugging.


zhangh43:
O0 instead of O3 also benefits debugging.
we'd better have a debug version makefile.(or use configure to set devel or prod version before make)

paul-guo-:
We've already had debug version options for clients.

-D_DEBUG_CLIENT -g -O0

PR #23, opened 2017-08-31
Fix pipeline using the new API

zhangh43:
The previous commit has introduce a new API to install and configure plcontainer, which is more user-friendly. We adopt these changes in the new test pipeline.
This PR also introduce custom openssl into plcontainer docker image due to the upstream artifacts change.


concourse/pipeline.yml
```diff
@@ -273,11 +301,13 @@ jobs:
   plan:
   - aggregate:
     - get: bin_gpdb_centos7
+    - get: gpdb_src
     - get: plcontainer_src
-      trigger: true
```
stanlyxiang: why not triggered by plcontainer_src like centos6 ?

management/bin/plcontainer
```diff
@@ -106,7 +106,9 @@ def remote_cmd(pool, cmd_str, hosts):
                         remoteHost = host.strip())
         pool.addCommand(cmd)
     pool.join()
+    results = pool.getCompletedItems()[0].get_results().stdout
```
stanlyxiang: This file contains changes of this file in PR #24 . And that PR #24 have some more commits, so I think this PR should resolve the conflicts by removing the changes or some other ways.

PR #27, opened 2017-09-08
Fix issues of function test on centos

zhangh43:
There are some issues of plcontainer function test
1 On centos7, Perl need to install cpan to support dumper which is needed by pg_regeress
2 On centos6, Docker version is too old, and there are no outputs for docker load command. So we add an option to specify the image name explicitly
3 Function test has false positive problem. We add "set -eox" in ssh command to fix it.

### DIFF
--- a/management/bin/plcontainer
+++ b/management/bin/plcontainer
@@ -65,7 +65,8 @@ def parseargs():
     # for install
     install_parser.add_argument("-i", "--image", dest="image", help="Configure container image location, support Tar.gz image" +
                         " file or docker image url")
-    install_parser.add_argument("-n", "--name", dest="name", help="Configure container image name")
+    install_parser.add_argument("-n", "--name", dest="name", help="Configure plcontainer type name")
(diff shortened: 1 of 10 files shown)


HaozhouWang:
+1

management/bin/plcontainer
```diff
@@ -65,7 +65,8 @@ def parseargs():
     # for install
     install_parser.add_argument("-i", "--image", dest="image", help="Configure container image location, support Tar.gz image" +
                         " file or docker image url")
-    install_parser.add_argument("-n", "--name", dest="name", help="Configure container image name")
+    install_parser.add_argument("-n", "--name", dest="name", help="Configure plcontainer type name")
+    install_parser.add_argument("-c", "--imagename", dest="imagename", help="Configure container image name")
```
paul-guo-: If this for old docker only, we should mention it in help. Others look good.

zhangh43: New docker version can also specify it, although it's not necessary.

PR #30, opened 2017-09-11
remove useless refresh since every sql file test in pg_regress will open new session

stanlyxiang:
Also, this will cause regression tests failed in a different cluster.


paul-guo-:
+1

zhangh43:
+1

PR #31, opened 2017-09-11
Delete containers when connection to container fails.

paul-guo-:
Else if client exits abnormally sql queries could continuously fail until QE terminates.

We do not apply the more friendly solution (e.g. seamlessly relaunch containers)
since it will introduce some complexity and this case is rare.
Instead we just tell users that query fails and then they could retry.

In this patch we also modify to not just kill containers, but
further delete containers.

### DIFF
--- a/tests/sql/plcontainer_function_python.sql
+++ b/tests/sql/plcontainer_function_python.sql
@@ -238,7 +238,7 @@ CREATE OR REPLACE FUNCTION pyreturnsetoftextarr(num int) RETURNS setof text[] AS
 # container: plc_python_shared
 def get_texts(n):
     return ['n'+str(x) for x in range(n)]
-    
+ 
 return [get_texts(x+1) for x in range(num)]
 $BODY$ LANGUAGE plcontainer;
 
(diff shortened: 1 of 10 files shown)


HaozhouWang:
+1

zhangh43:
what's the difference between kill and delete API?

paul-guo-:
@zhangh43 See https://docs.docker.com/engine/api/v1.24/
Basically it is the difference between "docker kill" and "kill rm". With delete you will not see it in "docker ps -a".

tests/sql/plcontainer_function_python.sql
```diff
@@ -567,3 +567,16 @@ CREATE OR REPLACE FUNCTION pyversion() RETURNS varchar AS $$
...
+import os
+import signal
+pid=os.getpid()
+os.kill(pid, signal.SIGKILL)
+$$ LANGUAGE plcontainer;
+
+CREATE OR REPLACE FUNCTION pyzero() RETURNS integer AS $$
+# container: plc_python_shared
+return 0
+$$ LANGUAGE plcontainer;
```
stanlyxiang: do we need to update out files as well?

zhangh43:
＋1

tests/sql/plcontainer_function_python.sql
paul-guo-: Yes. Forgot about this. Thanks.

PR #35, opened 2017-09-13
fix curl docker inspection funtion error

HaozhouWang:
This commit is fix curl docker inspect issue caused by cleanup commit.


paul-guo-:
+1

stanlyxiang:
+1

PR #37, opened 2017-09-14
support docker 1.7.1 (API v1.19) in Centos 6

HaozhouWang:
Add Centos 6 docker API support as its API version is too low and has been changed in newer version of docker.


stanlyxiang:
+1

paul-guo-:
+1

PR #38, opened 2017-09-15
Refactor docker API code.

paul-guo-:
1. Move some common docker api code to plc_docker_common.c
2. Do not elog(ERROR) in api code. Let callers decide how to proceed.
3. Add plc_docker_api_common.h
4. Fix some bugs in api code.
4. Various other refactor code.

We want to do this since we've found various issues due to this.


src/containers.c
```diff
@@ -81,10 +81,13 @@ static int container_is_alive(char *dockerid) {
...
         if (res < 0) {
             return_code = res;
-        }
-        if (element != NULL && (strcmp("exited", element) == 0 ||
-			strcmp("false", element) == 0 )){
-            return_code = plc_backend_delete(sockfd, dockerid);
+        } else if (element != NULL) {
+			if ((strcmp("exited", element) == 0 ||
+				strcmp("false", element) == 0 )) {
+				return_code = plc_backend_delete(sockfd, dockerid);
```
zhangh43: the return code of plc_backend_delete is defined as:
200 – no error
400 – bad parameter
500 – server error
we should treat 400 or 500 as failure.

zhangh43:
+1

paul-guo-:
Merged.

PR #39, opened 2017-09-18
Using new API to inject faults in pg_extension.

zhangh43:
Using SIMPLE_FAULT_NAME_INJECTOR instead of SIMPLE_FAULT_INJECTOR to inject faults by fault names.


paul-guo-:
+1

PR #41, opened 2017-09-18
Show diff output on concourse if testing fails.

paul-guo-:
This saves some time when there is test failure since usually users need to fly to concourse instance to check.


HaozhouWang:
+1

zhangh43:
the bash is run under "set -eox pipelinefail" mode. The script will return is one of the command fails.
I guess that the print diff command will not be executed.

paul-guo-:
It is in Makefile, so it works. I've tested it.

zhangh43:
+1

PR #42, opened 2017-09-19
Fix random failure of fault injection test case.

zhangh43:
We need to specify the exact name of "plcontainer cleaner" processes because in pipeline we introduce an additional process with name containing "plcontainer" when run test by bash "bash plcontainer.sh"


paul-guo-:
+1

stanlyxiang:
＋1

PR #44, opened 2017-09-20
Set flags on various points to remind container deletion.

paul-guo-:

### DIFF
--- a/src/plcontainer.c
+++ b/src/plcontainer.c
@@ -41,7 +41,7 @@ static void plcontainer_process_exception(plcMsgError *msg);
 static void plcontainer_process_sql(plcMsgSQL *msg, plcConn* conn);
 static void plcontainer_process_log(plcMsgLog *log);
 
-static bool DeleteBackendsRequired;
+static bool DeleteBackendsWhenError;
 
 Datum plcontainer_call_handler(PG_FUNCTION_ARGS) {
     Datum datumreturn = (Datum) 0;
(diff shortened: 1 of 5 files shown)


src/plcontainer.c
```diff
@@ -181,19 +181,22 @@ static plcProcResult *plcontainer_get_result(FunctionCallInfo  fcinfo,
             elog(ERROR, "Container '%s' is not defined in configuration "
                         "and cannot be used", name);
         } else {
-            conn = start_container(conf);
+			/* TODO: We could just remove this backend when error occurs. */
+			DeleteBackendsWhenError = true;
+            conn = start_backend(conf);
```
zhangh43: format need align.

src/plcontainer.c
```diff
@@ -220,6 +222,8 @@ static plcProcResult *plcontainer_get_result(FunctionCallInfo  fcinfo,
                     result->resrow = 0;
                     break;
                 case MT_EXCEPTION:
+					/* For exception, no need to delete containers. */
+					DeleteBackendsWhenError = false;
```
zhangh43: format problem too

paul-guo-: GitHub by default has the tab width as 8. You could use tab with as 4 in the code. (add ?ts=4 in url). Anyway I modified these indents (To use tab instead of space). Thanks.

tests/expected/plcontainer_faultinject_python.out
```diff
@@ -78,7 +93,7 @@ NOTICE:  Success:
 (1 row)
 
 SELECT pyint(i) from tbl;
-ERROR:  fault triggered, fault name:'plcontainer_after_send_request' fault type:'fatal'  (seg0 slice1 127.0.0.1:25432 pid=453)
+ERROR:  fault triggered, fault name:'plcontainer_after_send_request' fault type:'fatal'  (seg0 slice1 127.0.0.1:25432 pid=15442)
```
stanlyxiang: have we ignored this by init_file ? or we will change it every time.

HaozhouWang: Yep, this has been ignored.

zhangh43:
+1

PR #51, opened 2017-09-21
Rename some plpython functions/files to align with upstream.

paul-guo-:


zhangh43:
PLy_spi_execute is the internal function. User interface in python should still be plpy_execute()

zhangh43:
+1

PR #53, opened 2017-09-21
Retire docker socket API functions

HaozhouWang:
The curl API for docker is safer and easier to use.
This commit disable the docker socket API and set
curl API as default.

1. Update Makefile, now require libcurl version is > 7.40
2. Remove docker socket APIs
3. Remove connect and disconnect function in backend APIs to
   reduce the complexity of APIs


paul-guo-:
+1

PR #54, opened 2017-09-22
Refactor docker regex code to remove the duplicate code

HaozhouWang:
1. Combine get container id and inspection regex funtion into one
2. Remove conatiner id warnning regex, it is already logged into
   GPDB in previous log


paul-guo-:
+1

zhangh43:
+1

PR #63, opened 2017-09-28
Add release pipeline.

zhangh43:


concourse/scripts/build_plcontainer.sh
```diff
@@ -15,19 +15,24 @@ build_plcontainer() {
...
-      export CIBUILD=1
-  fi
-  
+ 
   # build plcontainer
   pushd plcontainer_src
-  make clean
+  if [ "${DEV_RELEASE}" == "release" ]; then
+      PLCONTAINER_VERSION=$(git describe | awk -F. '{printf("%d.%d", $1, $2)}')
+      PLCONTAINER_RELEASE=$(git describe | awk -F. '{print $3}')
```
paul-guo-: Maybe we should warn and exit if "git describe" is bad.

zhangh43: fixed it

PR #65, opened 2017-09-28
Add plpy function PLy_subtransaction

zhangh43:
This PR just add function PLy_subtransaction for plcontainer.
Enter and Exit of transaction is not implemented yet.


paul-guo-:
+1

PR #69, opened 2017-10-09
fix image tarball file not found

stanlyxiang:
the image tarball names are changed, so command scp cannot find the image tar.gz file. Use * instead.


paul-guo-:
+1

HaozhouWang:
+1

PR #86, opened 2017-10-17
Redirect python logs in container to host

zhangh43:
Also add more detailed DEBUG1 level logs in QE side to better debug the 404
error and timeout error.
Pipeline also fixed to use python on centos7 instead of centos6


src/containers.c
```diff
@@ -259,6 +260,12 @@ plcConn *start_backend(plcContainerConf *conf) {
 	 */
 	insert_container(conf->name, dockerid, container_slot);
 
+
+    time_t rawtime;
+    struct tm * timeinfo;
+    time ( &rawtime );
+    timeinfo = localtime ( &rawtime );
+    elog(DEBUG1, "container %s has started at %s", dockerid, asctime (timeinfo));
```
stanlyxiang: since container maybe started failed or need some time, we'd better move this line under ```res = plc_backend_start(dockerid);```.  Otherwise, it's confusing or inaccurate.

src/plcontainer.c
```diff
@@ -77,7 +77,7 @@ Datum plcontainer_call_handler(PG_FUNCTION_ARGS) {
         /* If the reason is Cancel or Termination or Backend error. */
         if (InterruptPending || QueryCancelPending || QueryFinishPending ||
 			DeleteBackendsWhenError) {
-            //elog(DEBUG1, "Terminating containers due to user request");
+            elog(DEBUG1, "Terminating containers due to user request");
```
stanlyxiang: we could print which condition leads to terminating containers for easier debug.

src/plc_docker_curl_api.c
```diff
@@ -214,7 +214,7 @@ int plc_docker_create_container(plcContainerConf *conf, char **name, int contain
 	} else if (res >= 0) {
 		elog(LOG, "Docker fails to create container, response: %s", response->data);
 		snprintf(api_error_message, sizeof(api_error_message),
-				"Failed to create container (return code: %d).", res);
+				"Failed to create container (return code: %d) response message %s.", res, response->data);
```
paul-guo-: api_error_message will be shown on psql output. We should not expose internal REST api details.

package/Makefile
```diff
@@ -16,7 +16,7 @@ PLC_DIR=`cd .. && pwd`
 PLC_GPPKG_VER=$(PLCONTAINER_VERSION).$(PLCONTAINER_RELEASE)
 PLC_RPM_FLAGS="--define 'plc_dir $(PLC_DIR)' --define 'plc_ver $(PLCONTAINER_VERSION)' --define 'plc_rel $(PLCONTAINER_RELEASE)'"
 PLC_RPM=plcontainer-$(PLCONTAINER_VERSION)-$(PLCONTAINER_RELEASE).$(ARCH).rpm
-PLC_GPPKG=plcontainer-$(PLC_GPPKG_VER)-rhel$(MAJOR_OS)-$(ARCH).gppkg
+PLC_GPPKG=plcontainer-$(PLC_GPPKG_VER)-gp5-rhel$(MAJOR_OS)-$(ARCH).gppkg
```
paul-guo-: Why not use GP_VERSION?

src/containers.c
```diff
@@ -356,7 +363,11 @@ void delete_containers() {
 
                 /* Terminate container process */
                 if (containers[i].dockerid != NULL) {
-                    plc_backend_delete(containers[i].dockerid);
+                    int res = plc_backend_delete(containers[i].dockerid);
+                    if (res < 0) {
+                        elog(LOG, "Failed to delete container in cleanup process (%s). "
```
paul-guo-: delete_containers() is not called in cleanup process so the comment is not accurate.

Both QE and cleanup process could delete a container so it is normal that plc_backend_delete() fails so no need of this code?

src/pyclient/client.c
```diff
@@ -32,7 +33,15 @@ int main(int argc UNUSED, char **argv UNUSED) {
...
 
-	sock = start_listener();
+    setbuf(stdout, NULL);
+    time_t rawtime;
+    struct tm * timeinfo;
+
+    sock = start_listener();
+    time ( &rawtime );
+    timeinfo = localtime ( &rawtime );
+    lprintf(LOG, "Client has started execution at %s", asctime (timeinfo));
```
paul-guo-: According to man page. Maybe we should use strftime()?

       POSIX.1-2001.  C89 and C99 specify asctime(), ctime(), gmtime(), localtime(), and mktime().  POSIX.1-2008 marks asctime(), asctime_r(),
       ctime(), and ctime_r() as obsolete, recommending the use of strftime(3) instead.

src/containers.c
zhangh43: yes, deleted it

package/Makefile
zhangh43: fixed

src/containers.c
zhangh43: fixed

src/plcontainer.c
zhangh43: added

HaozhouWang:
+1

paul-guo-:
+1

stanlyxiang:
+1

src/plcontainer.c
```diff
@@ -76,10 +76,12 @@ Datum plcontainer_call_handler(PG_FUNCTION_ARGS) {
     {
         /* If the reason is Cancel or Termination or Backend error. */
         if (InterruptPending || QueryCancelPending || QueryFinishPending ||
-			DeleteBackendsWhenError) {
-            //elog(DEBUG1, "Terminating containers due to user request");
+            DeleteBackendsWhenError) {
+            elog(DEBUG1, "Terminating containers due to user request reason(interrupt/cancle \
```
stanlyxiang: typo "cancle"

zhangh43: fixed

PR #89, opened 2017-10-18
Add index in the prefix of domain socket path.

zhangh43:
QE may correspond to many containers, and cleanup process will remove
the domain socket file asynchronously. We need to separate domain socket
file for different containers.

### DIFF
--- a/src/plc_configuration.c
+++ b/src/plc_configuration.c
@@ -23,6 +23,7 @@
 
 static plcContainerConf *plcContConf = NULL;
 static int plcNumContainers = 0;
+static int domain_socket_num = 0;
 
 static int parse_container(xmlNode *node, plcContainerConf *conf);
 static plcContainerConf *get_containers(xmlNode *node, int *size);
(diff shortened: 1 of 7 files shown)


src/plc_configuration.c
```diff
@@ -23,6 +23,9 @@
 
 static plcContainerConf *plcContConf = NULL;
 static int plcNumContainers = 0;
+// we just want to avoid cleanup process to remove previous domain
+// socket file, so int32 is sufficient
```
paul-guo-: Better use /* for comment.

src/plc_configuration.c
```diff
@@ -407,20 +410,19 @@ char *get_sharing_options(plcContainerConf *conf, int container_slot, bool *has_
                 comma = ',';
 			/* Directory for QE : IPC_GPDB_BASE_DIR + "." + PID + "." + container_slot */
```
paul-guo-: Comment is wrong now.

PR #92, opened 2017-10-18
enhance docker curl api in function plcCurlRESTAPICall

stanlyxiang:


HaozhouWang:
+1

src/plc_docker_curl_api.c
```diff
@@ -80,13 +80,15 @@ static plcCurlBuffer *plcCurlRESTAPICall(plcCurlCallType cType,
 
     memset(errbuf, 0, CURL_ERROR_SIZE);
 
+    curl_global_init(CURL_GLOBAL_ALL);
```
zhangh43: should not global init here. It has already init in postgres.c

src/plc_docker_curl_api.c
```diff
@@ -148,17 +150,20 @@ static plcCurlBuffer *plcCurlRESTAPICall(plcCurlCallType cType,
...
         pfree(fullurl);
-		curl_easy_cleanup(curl);
+        curl_slist_free_all(headers);
+        curl_easy_cleanup(curl);
     } else {
 		snprintf(api_error_message, sizeof(api_error_message),
 				"Failed to start a curl session for unknown reason");
 		buffer->status = -1;
 	}
+    curl_global_cleanup();
```
zhangh43: don't needed neither

zhangh43:
Put global init in postgres.c maybe a waste when user don't use plcontainer at all.
But global_init for each libcurl call in plcontainer udf is also a waste.
+1 for this PR, but we need to pay attention to the performance introduced by global_init and global_cleanup for each udf.

PR #94, opened 2017-10-19
Add mask parameter into plcontainer_channel_receive().

paul-guo-:
With this, we can filter unexpected messages earlier.
Besides, we could easiler free for unexpected message if needed in this function.


HaozhouWang:
+1

PR #95, opened 2017-10-20
Quicly fix pipline failure due to rclient compile warning(error).

paul-guo-:


paul-guo-:
Easy fix. Will check in directly since this is blocking the pipeline.

PR #103, opened 2017-10-24
Fix SPI function to support INSERT, UPDATE and DELETE

HaozhouWang:
1. Fix backend to support non-returning INSERT, UPDATE and DELETE
2. Fix python client to support non-returning INSERT, UPDATE and DELETE
   and will return the number of rows that are processed.

### DIFF
--- a/src/common/comm_messages.c
+++ b/src/common/comm_messages.c
@@ -127,9 +127,16 @@ void free_result(plcMsgResult *res, bool isSender) {
         }
         free_type(&res->types[i]);
     }
-    pfree(res->types);
-    pfree(res->names);
 
+	if (res->types != NULL){
+		pfree(res->types);
+		res->types = NULL;
+	}
+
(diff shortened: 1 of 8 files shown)


src/common/comm_messages.c
```diff
@@ -127,9 +127,14 @@ void free_result(plcMsgResult *res, bool isSender) {
         }
         free_type(&res->types[i]);
     }
-    pfree(res->types);
-    pfree(res->names);
 
+	if (res->types != NULL){
+		pfree(res->types);
```
zhangh43: better also set res->types=NULL

paul-guo-:
+1

src/common/comm_messages.c
HaozhouWang: Thanks, fixed

zhangh43:
+1

PR #108, opened 2017-10-25
Detect cleanup process name change.

paul-guo-:
If cleanup process name changes, fault injection tests seem to still succeed,
but fail to detect cleanup related failures.


HaozhouWang:
+1

PR #113, opened 2017-10-25
Fix plcontainer install to add language option.

zhangh43:
This bug was not found before due we replace conf file by a template file.
I remove replace code in performance test
```
 plcontainer configure -f plcontainer_src/tests/plcontainer_configuration_test.xml -y; 
```
 which could test whether language option works.


HaozhouWang:
+1

PR #114, opened 2017-10-25
Clean unused test files and enable error tests

HaozhouWang:
1. Remove unused Python/R function tests
2. Enable Python/R error test cases


paul-guo-:
Why do we delete tests/sql/plcontainer_endurance_functions_gpdb5.sql and
tests/sql/plcontainer_endurance_gpdb5.sql?

Aren't they are useful, although they are not enabled?

paul-guo-:
+1

PR #120, opened 2017-10-27
Enhance [de]serialization (unify type width; sanity check; rx timeout).

paul-guo-:


zhangh43:
+1

PR #132, opened 2017-11-03
Refactor test Makefile and add memory stress tests

HaozhouWang:
1. Refactor test Makefile, now using schedule file for regression
   tests.
2. Add memory stress tests for both Python and R


paul-guo-:
I'm wondering we should remove the plcontainer prefix for all (including previous) test files?

zhangh43:
+1

PR #146, opened 2017-11-10
Delete containers in on_proc_exit of QE also.

paul-guo-:
We saw a sigsegv issue in the cleanup process and thus finally there are
containers which are not cleaned up. This reminds us that we better
delete containers also in on_proc_exit.

This patch also adds dockerid information in cleanup code.

### DIFF
--- a/src/containers.c
+++ b/src/containers.c
@@ -83,7 +83,6 @@ static int container_is_alive(char *dockerid) {
     }
 
     return return_code;
-
 }
 
 static void cleanup4uds(char *uds_fn) {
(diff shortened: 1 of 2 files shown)


src/containers.c
```diff
@@ -197,12 +196,15 @@ static void cleanup(char *dockerid, char *uds_fn) {
...
 
+			elog(LOG, "cleanup process deleted docker %s with return value %d",
+				 dockerid, res);
 			exit(res);
 		}
 		PG_CATCH();
 		{
 			/* Do not rethrow to previous stack context. exit immediately.*/
-			elog(LOG, "cleanup process should not reach here. Anyway it should not hurt. Exiting.");
+			elog(LOG, "cleanup process should not reach here. Anyway it should"
```
zhangh43: how about add message:
May lead to container leak, please run 'docker ps -a' to check and remove the container manually.

paul-guo-: Agree

PR #149, opened 2017-11-11
Do not call curl_global_init() multiple times althouth upstream code …

paul-guo-:
…seems to check this itself.

### DIFF
--- a/src/plc_docker_curl_api.c
+++ b/src/plc_docker_curl_api.c
@@ -81,17 +81,11 @@ static plcCurlBuffer *plcCurlRESTAPICall(plcCurlCallType cType,
 
     memset(errbuf, 0, CURL_ERROR_SIZE);
 
-    /* FIXME:
-	 * pay attention to the performance introduced by global init and cleanup
-     * also a new curl handle for each udf is much more waste. need further 
-     * performance optimization.
-     */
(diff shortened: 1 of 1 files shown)


stanlyxiang:
+1

src/plc_docker_curl_api.c
```diff
@@ -78,20 +78,22 @@ static plcCurlBuffer *plcCurlRESTAPICall(plcCurlCallType cType,
...
-     * also a new curl handle for each udf is much more waste. need further 
-     * performance optimization.
-     */
-    curl_global_init(CURL_GLOBAL_ALL);
+	if (initialized == 0) {
+		/* We might want to add common interfaces in the future so that
+		 * curl_global_cleanup() could be called (e.g. in atexit callback).
+		 */
+		curl_global_init(CURL_GLOBAL_ALL);
+		initialized = 1;
```
HaozhouWang: We do not need to call curl_global_init() explicit as curl_easy_init() has did this job. 
Ref: https://github.com/curl/curl/blob/aa7668b94815fd880faecaa4bc4a9ed59e07435c/lib/easy.c#L344

paul-guo-:
Merged. Closing. Thanks @stanlyxiang @HaozhouWang

PR #157, opened 2017-11-18
Eliminate the security concern for unix domain socket connection.

paul-guo-:
Previously the client could write under the directory for unix domain socket file sharing.
Now we change the permission and ownership of the file after it is created and then
downgrade from root to usual user that can not write the path.


src/common/comm_server.c
```diff
@@ -89,52 +91,61 @@ static int start_listener_ipc(char **puds_fn) {
...
-
-	/* Save it since we need to unlink the file when container process exits. */
-	if (puds_fn != NULL) {
-		unlink(*puds_fn);
-		pfree(*puds_fn);
-	}
-	*puds_fn = uds_fn;
+	if((env_str = getenv("EXECUTOR_UID")) == NULL)
+		lprintf(ERROR, "EXECUTOR_UID is not set, something wrong on QE side");
+	uid = strtol(env_str, NULL, 10);
```
HaozhouWang: Do we need to validate the input string if it is not validated?

paul-guo-: This is a nice-to-have check. OK I'll add them. Thanks.

src/containers.c
```diff
@@ -372,16 +376,14 @@ plcConn *start_backend(plcContainerConf *conf) {
         char *element = NULL;
 		res = plc_backend_inspect(dockerid, &element, PLC_INSPECT_PORT);
 		if (res < 0) {
+			cleanup4uds(uds_fn);
```
zhangh43: elog(ERROR) will be catched by portal function, and cleanup4uds will be called in the catch logic. Do we need to call it here?

paul-guo-: delete_containers() will not cleanup by now. We might want to include in the future?

PR #161, opened 2017-11-22
Fix cleanup process longjump to postgres main function

zhangh43:
It may leads to double unpin the refcount of buffer. The RCA is the
use of elog outside the try-catch block of cleanup process, we now
use write_log instead.

This patch also add DEBUG1 log_min_messages in faultinjector test.

Signed-off-by: Paul Guo <paulguo@gmail.com>


wuj21:
+1

PR #171, opened 2017-11-24
The original license needs to be added back to the project. Also corr…

davecramer:
…ect attribution for plcontainer.c to Lazlo

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,5 @@
-Copyright 1994-2004 The PL-J Project. All rights reserved.
+Portions Copyright 1994-2004 The PL-J Project. All rights reserved.
+Portions Copyright © 2016-Present Pivotal Software, Inc.
 
 Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
 
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,10 @@
+Copyright 1994-2004 The PL-J Project. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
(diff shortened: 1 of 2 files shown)


paul-guo-:
@davecramer  Please do not remove the Pivotal licence header since lots of changes in the file are done by Pivotal employees. And is plpgj_core.c
https://github.com/codehaus/plj/blob/master/pl-j/csrc/plpgj_core.c? Assuming yes, do you have more context about why this information should be included given it seems that that file just includes the very common code which is used in other PL language code also. Thanks.

davecramer:
@paul-guo- yes, that is the original file, and since you can't simply remove previous authorship it needs to be included. Unfortunately at the time I was a bit too aggressive when I removed the original authorship. Whether or not you can later copyright that file or not is up to the lawyers

schubert:
@paul-guo- @yydzero I'll take this PR up. Let me take a look at it and I'll let you all know how to proceed.

yydzero:
Thanks @schubert !

davecramer:
Seems reasonable to me that both should be "Portions Copyright"

Dave Cramer

LICENSE.txt
```diff
@@ -0,0 +1,10 @@
+Copyright 1994-2004 The PL-J Project. All rights reserved.
```
schubert: Should be:

```
Portions Copyright 1994-2004 The PL-J Project. All rights reserved.
Portions Copyright © 2016-Present Pivotal Software, Inc.
```

paul-guo-:
src/plcontainer.c should include Pivotal information also, right?

@schubert @davecramer

davecramer:
@paul-guo- sure, makes sense.

paul-guo-:
@davecramer  Please modify src/plcontainer.c and then it will be checked in.

Issue #88, opened 2017-10-18
Reformat the code to use tab for indent.


paul-guo-:
To align with upstream (greenplum and postgres).

HaozhouWang:
Fixed in this PR:
https://github.com/greenplum-db/plcontainer/pull/187

PR #187, opened 2017-12-04
Reformat the source code

HaozhouWang:
Using tab to instead of space


HaozhouWang:
Related to https://github.com/greenplum-db/plcontainer/issues/88

PR #235, opened 2017-12-19
Add python client SPIError

HaozhouWang:
1. Add SPIError exception for PL/Container python client
2. Now plpy SPI interface will throw exception and could be caught in PL Python code.
3. Add plpy.SPIError as new SPI exception object in python client.


zhangh43:
try this case 
```
CREATE FUNCTION transfer_funds2() RETURNS void AS $$
try:
    with plpy.subtransaction():
        plpy.execute("UPDATE accounts SET balance = balance - 100 WHERE account_name = 'joe'")
        plpy.execute("UPDATE accounts SET balance = balance + 100 WHERE account_name = 'mary'")
except plpy.SPIError, e:
    result = "error transferring funds: %s" % e.args
else:
    result = "funds transferred correctly"
plan = plpy.prepare("INSERT INTO operations (result) VALUES ($1)", ["text"])
plpy.execute(plan, [result])
$$ LANGUAGE plpythonu;
```

src/pyclient/plpy_spi.c
```diff
@@ -407,13 +438,13 @@ PLy_spi_execute_plan(PyObject *ob, PyObject *list, long limit) {
 	 * so if resp->cols > 0, it must be SELECT statment.
 	 */
 	if (resp->cols == 0) {
-		debug_print(WARNING, "the rows is %d", resp->rows);
+		debug_print(NOTICE, "the rows is %d", resp->rows);
```
paul-guo-: DEBUG1? debug_print->lprintf?

src/pyclient/plpy_spi.c
```diff
@@ -685,13 +718,13 @@ PLy_spi_execute_query(char *query, long limit) {
...
```
paul-guo-: See previous comment.

tests/sql/spi_python.sql
```diff
@@ -371,3 +371,16 @@ select pyspi_delete_notexist_execp();
...
+CREATE OR REPLACE FUNCTION pyspi_exec_exception() RETURNS integer AS $$
+# container: plc_python_shared
+plan = list();
+try:
+	rv = plpy.execute(plan);
+except Exception as e:
+	plpy.notice(type(e))
+return 0
+$$ LANGUAGE plcontainer;
+
```
paul-guo-: Need more test cases to cover code branches.

HaozhouWang:
Tested this case, the error could not be caught by client as the type conversion error has been caught by QE side and not passed to client.

tests/sql/spi_python.sql
HaozhouWang: Thx, fixed

PR #236, opened 2017-12-19
xml file validation code in plcontainer utility should be more aggressive

paul-guo-:


management/bin/plcontainer
```diff
@@ -435,19 +436,96 @@ def copy_file_to_tmp(host, file):
...
+
+            # Only one 'image' subelement is allowed.
+            num = len(runtime.findall('image'))
+            if num < 1:
+                logger.error("Lack of 'image' subelement in a runtime element in runtime %s.", runtime_id)
+                raise Exception("Validation failed")
+            if num > 1:
+                logger.error("Found %d (1 is legal) 'image' subelements in a runtime %s. The first one 'image': %s", num, runtime_id, runtime.find('image').text)
+                raise Exception("Validation failed")
+            # FIXME: Need to check the naming convention?
```
zhangh43: how about check image is loaded on docker ?

management/bin/plcontainer
```diff
@@ -435,19 +436,96 @@ def copy_file_to_tmp(host, file):
...
+                if 'access' in share_folder.attrib and share_folder.attrib['access'] != 'ro' and share_folder.attrib['access'] != 'rw':
+                    logger.error("access should be 'ro' or 'rw' in shared_directory in runtime %s, but now: %s", runtime_id, share_folder.attrib['access'])
+                    raise Exception("Validation failed")
+
+                if  share_folder.attrib.keys() != ['access', 'host', 'container']:
+                    logger.error("Unknown or missing attributes (need and only need 'access', 'host' and 'container') in shared_directory in runtime %s", runtime_id)
+                    logger.error("Current: %s", share_folder.attrib.keys())
+                    raise Exception("Validation failed")
+                container_pathlist.append(share_folder.attrib['container'])
+
```
zhangh43: how about duplicate host_pathlist?

management/bin/plcontainer
paul-guo-: Yes, we could check this in runtime-add, runtime-replace, and this function. Maybe we could defer this to another issue.

management/bin/plcontainer
paul-guo-: dup host paths are fine.

zhangh43:
+1

PR #241, opened 2017-12-26
Make some gpssh related outputs in verbose output to avoid too many l…

paul-guo-:
…ogs for large cluster.

### DIFF
--- a/management/bin/plcontainer
+++ b/management/bin/plcontainer
@@ -560,7 +560,7 @@ def distribute_file(pool, src_filename, dest_filename, gpdb_locations):
     """
     for l in gpdb_locations:
         dest_file = os.path.join(l[1], dest_filename)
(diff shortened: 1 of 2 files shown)


management/bin/plcontainer
```diff
@@ -804,7 +804,7 @@ def main():
             update_xml_config(tmp_file, elements)
             logger.debug('Preparing the worker pool...')
             pool = WorkerPool(PL_BATCH_SIZE)
-            logger.info('Distributing file %s to all locations...' % PLCONTAINER_CONFIG_FILE)
+            logger.debug('Distributing file %s to all locations...' % PLCONTAINER_CONFIG_FILE)
```
zhangh43: better keep this message.

paul-guo-: Agree.

PR #257, opened 2018-01-04
Enhance logging in plcontainer

paul-guo-:
Use plc_elog() instead of elog() so that we could have seperate logging control than gpdb in the future.
Replace debug_print() with more meaningful function name channel_elog().
Replace all lprintf() with more meaning name plc_elog() also.

### DIFF
--- a/src/common/comm_channel.c
+++ b/src/common/comm_channel.c
@@ -98,7 +98,7 @@ static int receive_rawmsg(plcConn *conn, plcMessage **mRaw);
 
 int plcontainer_channel_send(plcConn *conn, plcMessage *msg) {
 	int res;
-	lprintf(DEBUG1, "start to send data, type is %c", msg->msgtype);
+	plc_elog(DEBUG1, "start to send data, type is %c", msg->msgtype);
 	switch (msg->msgtype) {
 		case MT_PING:
 			res = send_ping(conn);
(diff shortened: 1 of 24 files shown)


paul-guo-:
@zhangh43 @HaozhouWang

paul-guo-:
rclient code change will be added after the change enters.

zhangh43:
plc_elog will add prefix on QE side, how about client side?

paul-guo-:
@zhangh43  See common/comm_utils.h
```
#define plc_elog(lvl, fmt, ...)                                          \
        do {                                                            \
            FILE *out = stdout;                                         \
            if (lvl >= ERROR) {                                         \
                out = stderr;                                           \
            }                                                           \
            fprintf(out, "plcontainer log: %s, ", clientLanguage);      \
            fprintf(out, "%s, %s, %d, ", dbUsername, dbName, dbQePid);  \
            fprintf(out, #lvl ": ");                                    \
            fprintf(out, fmt, ##__VA_ARGS__);                           \
            fprintf(out, "\n");                                         \
            fflush(out);                                                \
            if (lvl >= ERROR) {                                         \
                exit(1);                                                \
            }                                                           \
        } while (0)

```

zhangh43:
+1

PR #263, opened 2018-01-10
Add Pivotal license information since those files were modified by Pi…

paul-guo-:
…votal employees.

Author: Haozhou Wang <hawang@pivotal.io>
Author: Paul Guo <paulguo@gmail.com>


paul-guo-:
@schubert @davecramer 

We are adding some missing Pivotal license information in some files. Please review. Thanks.

paul-guo-:
According to the git change log, all the files in this patch have contributions from Pivotal since 2016. So the patch should be ok. I'm checking in the code.

PR #275, opened 2018-01-12
Add coverage in Makefile and fix compile warning.

zhangh43:


Makefile
```diff
@@ -22,6 +22,8 @@ OBJS = $(foreach FILE,$(FILES),$(subst .c,.o,$(FILE)))
 PGXS := $(shell pg_config --pgxs)
 include $(PGXS)
 
+CLIENT_CFLAG = -Werror -Wextra -Wall
```
paul-guo-: Why we do this change?

Makefile
```diff
@@ -58,13 +60,14 @@ endif
 PLCONTAINERDIR = $(DESTDIR)$(datadir)/plcontainer
 
 override CFLAGS += -Werror -Wextra -Wall
+override CFLAGS += $(COVERAGE_FLAG)
+override SHLIB_LINK += $(COVERAGE_LINK_FLAG)
```
HaozhouWang: Will we need to re-build GPDB for coverage flag?

zhangh43: we use gpdb extension(PGXS := $(shell pg_config --pgxs) include $(PGXS)) to build plcontainer.
And to enable code coverage, we need add compile and link flag.

Makefile
zhangh43: Client should not inherit the flag of plcontainer.so.
In future we may enable code coverage on client side

Makefile
HaozhouWang: Sure. Thanks

PR #280, opened 2018-01-25
Enforce to check if backend api code return value is not used.

paul-guo-:
We should do that so that the code could exit earlier and report more precise error.
This is inspired by the fix in my patch below.

commit a38fd5ca751bd3daf71a3a7ab3d375ef05d142ee
Error messages are missing in some docker api code. (#279)


zhangh43:
+1

PR #286, opened 2018-01-31
Combine plcontainer resource management with  gpdb resource group.

zhangh43:
This PR reset the cgroup parent of containers from 'docker' to a cgroup node under gpdb.
All the memory used in plcontainer can be managed by gpdb resource group.

### DIFF
--- a/src/plc_docker_api.c
+++ b/src/plc_docker_api.c
@@ -228,6 +228,7 @@ int plc_docker_create_container(runtimeConfEntry *conf, char **name, int contain
 			"    \"Image\": \"%s\",\n"
 			"    \"HostConfig\": {\n"
 			"        \"Binds\": [%s],\n"
+			"        \"CgroupParent\": \"%s\",\n"
 			"        \"Memory\": %lld,\n"
 			"        \"PublishAllPorts\": true,\n"
 			"        \"LogConfig\":{\"Type\": \"%s\"}\n"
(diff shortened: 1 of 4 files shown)


src/plc_docker_api.c
```diff
@@ -269,6 +270,16 @@ int plc_docker_create_container(runtimeConfEntry *conf, char **name, int contain
...
+	/*
+	 * generate cgroup parent path when using resource group feature
+	 * set cgroupParent to empty string when resgroupOid is not valid
+	 * and cgroup parent of containers will be 'docker' as default.
+	 * Note that this feature is only for gpdb with resource group enable.
+	 */
+	char cgroupParent[RES_GROUP_PATH_MAX_LENGTH] = "";
+	if (conf->resgroupOid != InvalidOid) {
+		snprintf(cgroupParent,RES_GROUP_PATH_MAX_LENGTH,"/gpdb/%d",conf->resgroupOid);
+	}
```
paul-guo-: Do we need RES_GROUP_PATH_MAX_LENGTH? palloc?

zhangh43: cgroupParent is just a temp variable, and when resgroup is not used, it should be an empty string. I think palloc and char array are both OK.

zhangh43:
Add three hooks in the second commits.
Hook1 free resource in container cgroup to gpdb memory pool
Hook2 get memory from gpdb memory pool to enlarge memory limit container cgroup.
Hook3 cleanup container when memory usage is high.

PR #287, opened 2018-01-31
Replace COMM_STANDALONE with more understandable PLC_CLIENT

paul-guo-:
Also modify README.md a bit.


zhangh43:
@paul-guo-  please also change the macro in rclient repo.

paul-guo-:
@zhangh43  Sure.

PR #298, opened 2018-02-08
Fix various issues in README.md

paul-guo-:


README.md
```diff
@@ -75,18 +72,9 @@ There are a couple of things you need to pay attention to:
...
-interpreter, simply make the following definition:
-```sql
-CREATE FUNCTION dummyR() RETURNS text AS $$
-# container: plc_r_shared
-return (log10(100))
-$$ LANGUAGE plcontainer;
-```
+1. The function definition starts with the line `# container: plc_python_shared` which defines the name of runtime that will be used for running this function. To check the list of runtimes defined in the system you can run the command `plcontainer runtime-show`. Each runtime is mapped to a single docker image, you can list the ones available in your system with command `docker images`
 
+plcontainer supports various parameters for docker run, and also it supports some useful UDFs for monitoring or debugging. Please read the official document for details.
```
zhangh43: here the debugging means enable log?

paul-guo-: e.g. plcontainer_containers_summary

PR #309, opened 2018-02-11
Wait for several seconds to give client a chance to exit normally ins…

paul-guo-:
…tead of being killed with SIGKILL.

This is needed when collecting code coverage report for client code.

Also rename api_error_message to backend_error_message which is more explicit.

### DIFF
--- a/src/containers.c
+++ b/src/containers.c
@@ -66,7 +66,7 @@ static int qe_is_alive(char *dockerid) {
 	return return_code;
 }
 
-static int container_is_alive(char *dockerid) {
+static int delete_backend_if_exited(char *dockerid) {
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -125,6 +125,10 @@ clean-coverage:
 	rm -f `find . -name '*.gcno' -print`
 	rm -rf coverage.info coverage_result
 
+.PHONY: clean-misc
+clean-misc:
+	rm -f management/sql/plcontainer_install.sql management/sql/plcontainer_uninstall.sql
+
```
HaozhouWang: we do not need this, *.sql will be automatically cleaned, when run ```make clean```

PR #311, opened 2018-02-11
Add test for plpy.prepare().status() and plpy.fatal().

zhangh43:


tests/sql/spi_python.sql
```diff
@@ -517,3 +518,10 @@ select nested_call_one('pass this along');
 select spi_prepared_plan_test_one('doe');
 select spi_prepared_plan_test_one('smith');
 select spi_prepared_plan_test_nested('smith');
+
+CREATE FUNCTION spi_fatal() RETURNS void AS $$
+# container: plc_python_shared
+plpy.fatal("test plpy fatal")
+$$ LANGUAGE plcontainer;
```
paul-guo-: We should move this to test_python.sql as other similar functions are there. Refer pylogging().

PR #321, opened 2018-02-26
Add cpu.share setting for runtime configuration

zhangh43:
We want to make containers of different runtime types using different percentage of CPU. This could be done through tuning the cpu.share of cgroup node.
'docker create' command support using cpu.share property to tuning the cpu.share value of cgroup node for this container.  This PR follows this idea to support this feature.

### DIFF
--- a/src/plc_configuration.c
+++ b/src/plc_configuration.c
@@ -411,7 +411,7 @@ static void print_runtime_configurations() {
 			plc_elog(INFO, "Container '%s' configuration", conf_entry->runtimeid);
 			plc_elog(INFO, "    image = '%s'", conf_entry->image);
 			plc_elog(INFO, "    memory_mb = '%d'", conf_entry->memoryMb);
-			plc_elog(INFO, "    cpuShare = '%d'", conf_entry->cpuShare);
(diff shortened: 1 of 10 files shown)


src/plc_configuration.c
```diff
@@ -398,6 +411,7 @@ static void print_runtime_configurations() {
 			plc_elog(INFO, "Container '%s' configuration", conf_entry->runtimeid);
 			plc_elog(INFO, "    image = '%s'", conf_entry->image);
 			plc_elog(INFO, "    memory_mb = '%d'", conf_entry->memoryMb);
+			plc_elog(INFO, "    cpuShare = '%d'", conf_entry->cpuShare);
```
HaozhouWang: Could we add an unit for cpuShare like percentage to help user to understand the meaning of the value of the cpuShare?

zhangh43: cgroup cannot set the cpu percentage, it can only set the relative value of cpu share. will document it.

paul-guo-:
Please modify management/config/plcontainer_configuration.xml also.

Others look good.

PR #333, opened 2018-03-12
Fix SearchSysCache does not use correctly

HaozhouWang:
In plc_typeio.c, SearchSysCache() inputs is datum, but
the current input is Oid. This commit fix this bug.


zhangh43:
+1

PR #335, opened 2018-03-20
Add Notification when container is OOM killed

HaozhouWang:
1. Send a warning message to user if a container is
   out of memory and killed by docker oom killer
2. If this case is caught by cleanup process, a log
   will be written
3. Add related test cases

Co-authored-by: Shaoqi Bai(baishaoqicoding@gmail.com)
Co-authored-by: Haozhou Wang(hawang@pivotal.io)

### DIFF
--- a/src/containers.c
+++ b/src/containers.c
@@ -66,11 +66,39 @@ static int qe_is_alive(char *dockerid) {
 	return return_code;
 }
 
+static int check_container_if_oomkilled(char *dockerid) {
+	char *element = NULL;
+
+	int return_code;
+	int res;
+
+	/*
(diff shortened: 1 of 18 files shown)


src/containers.c
```diff
@@ -66,19 +66,60 @@ static int qe_is_alive(char *dockerid) {
 	return return_code;
 }
 
+static int check_container_if_oomkilled(char *dockerid) {
+	char *element = NULL;
+
+	int return_code = 2;
```
zhangh43: better to replace magic number by meaningful name

HaozhouWang: Thanks, updated.

PR #355, opened 2018-04-26
Fix release pipeline error in GCP

HaozhouWang:
1. Fix image_build job in GCP concourse.
2. Change to `git describe --tags` to get correct tags

No need to test as this commit is related to concourse pipeline


BaiShaoqi:
+1

Issue #359, opened 2018-05-18
Coredump is caused by elog in cleanup process


zhangh43:
We should not use elog in cleanup process under resource group mode.
Since resource group has the assumption that forked subprocess should not 
share/use the resource group slot at the same time

paul-guo-:
Maybe we should create an issue now about using bg worker (pg 9.3 functionality) to do cleanup as some of us discussed? PG 9.3 functionatlies should be in gpdb master in a couple of months.

PR #360, opened 2018-05-18
Remove elog in cleanup process.

zhangh43:
## Description
palloc may be called inside elog, but QE process
and cleanup process share the same resource group
slot to do memory auditor. This slot maybe freed
after QE process exits, which cause elog of cleanup
process to corrupt. We need to use write_log to replace
elog in cleanup process.

## Tests

The tests are already exists, make them green under resource group mode.
## User Story or Issue

#359

Co-authored-by: Haozhou Wang <hawang@pivotal.io>

### DIFF
--- a/src/common/comm_utils.h
+++ b/src/common/comm_utils.h
@@ -122,6 +122,18 @@ int sanity_check_client(void);
 
 #include "postgres.h"
 
+/* QE process uses elog, while cleanup process should use write_log*/
+#define backend_log(elevel, ...)  \
+	do { \
+		if (getppid() == PostmasterPid) { \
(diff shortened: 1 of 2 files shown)


paul-guo-:
Maybe backend_elog()? Since for non-cleanup processes this function is more than logging (but also possibly exception catch, panic, etc).

zhangh43:
backend_log may also call write_log, which only support logging.(For case of cleanup process)
So backend_log is a better name which tells it may call write_log in non QE process.

zhangh43:
If the codes are only called by QE, we recommend to use the old plc_elog, which could also handle ERROR except just LOG

PR #373, opened 2018-05-29
Support User Access Control White List For Selected Runtimes

HaozhouWang:
## Description

1. Each runtime now support user white list, once set,
   a user could use this runtime to run UDF only if this
   user is in white list.
2. A new setting attribute "users" is used for white list.
   For example,
   ```
   <users>gpadmin</users>
   ```
   can only allow gpadmin user/role to use this runtime.
3. Add related regression tests

## Tests

Regression tests added

### DIFF
--- a/src/plc_configuration.c
+++ b/src/plc_configuration.c
@@ -166,7 +166,7 @@ static void parse_runtime_configuration(xmlNode *node) {
 		conf_entry->useContainerNetwork = false;
 		conf_entry->resgroupOid = InvalidOid;
 		conf_entry->useUserControl = false;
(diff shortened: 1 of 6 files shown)


src/plc_configuration.c
```diff
@@ -846,3 +865,34 @@ containers_summary(pg_attribute_unused() PG_FUNCTION_ARGS) {
 	}
 
 }
+
+bool plc_check_user_privilege(char *users){
```
zhangh43: We also need to consider group here. e.g. using is_member_of_role

Issue #423, opened 2018-06-08
List unsupported feature in README.md


zhangh43:
PLContainer is not fully compatible with plpython and plr yet, we have some features not supported. 
Need to list them in the README.md.

PR #424, opened 2018-06-08
Add unsupported feature link in README.md

zhangh43:
## Description

details refer to wiki link.
## Tests

only update readme.
## User Story or Issue

#423


HaozhouWang:
Good Catch!

Issue #381, opened 2018-06-01
ERROR when execute fucntion with default param and out param


BaiShaoqi:
### Expected behavior

 ```
 multiout_simple_setof
 -----------------------
  (1,2)
 (1 row)

```
### Actual behavior
```
 FATAL:  plcontainer: something bad happened, nargs (1) != len (3) (message_fns.c:139)
 HINT:  Process 21543 will wait for gp_debug_linger=120 seconds before termination.
 Note that its locks and other resources will not be released until then.
 server closed the connection unexpectedly
 	This probably means the server terminated abnormally
 	before or while processing the request.
 The connection to the server was lost. Attempting reset: Succeeded.
```
### Step to reproduce the behavior
```
CREATE FUNCTION multiout_simple_setof(n integer = 1, OUT integer, OUT integer) RETURNS SETOF record AS $$
# container: plc_python_shared
return [(1, 2)]*n
$$ language plcontainer;

SELECT multiout_simple_setof();
```

**Comment by zhangh43** [comment]
This is due to when we parse input"n integer = 1, OUT integer, OUT integer", we will get the func_arg_name (n,"","") which include three elements, but in fact only the first parameter is of input type.

**Comment by zhangh43** [comment]
What are OUT parameters? These are parameters you define as part of the function argument list that get returned back as part of the result. When you create functions, the arguments are defaulted to IN parameters when not explicitly specified 

One of the common use cases for using OUT parameters is to be able to return multiple outputs from a function without having to declare a PostgreSQL type as output of the function. In this article we shall cover all variants of this.

Another example :
```
CREATE OR REPLACE FUNCTION fn_sqltestout(param_subject text, 
    OUT subject_scramble text, OUT subject_char text)
   AS
$$
    SELECT  substring($1, 1,CAST(random()*length($1) As integer)), substring($1, 1,1)
    $$
  LANGUAGE 'sql' VOLATILE;
  
SELECT  (fn_sqltestout('This is a test subject')).subject_scramble;
```

PR #425, opened 2018-06-11
Fix arg number mismatch caused by OUT parameter.

zhangh43:
## Description

Argument length of function in catalog will not count OUT parameter in 'pronargs' field.
## Tests

Test added 
## User Story or Issue

#381


BaiShaoqi:
Can this pr pass the following test?

CREATE FUNCTION multiout_simple_setof(n integer = 1, OUT integer, OUT integer) RETURNS SETOF record AS $$
# container: plc_python_shared
return [(1, 2)]*n
$$ language plcontainer;

SELECT multiout_simple_setof();

PR #450, opened 2018-08-07
Change procpid to pid due to catalog change

HaozhouWang:
## Description

change procpid to pid due to catalog change
## Tests

No new test need


paul-guo-:
Does not this affect that on GP5?

HaozhouWang:
Thanks @paul-guo- , yes, it will be. I will update this patch to make two version for both gpdb 5 and master.

HaozhouWang:
Fixed, thanks for review.

PR #470, opened 2019-04-03
Comsuming binary from gpdb_master folder on gcs.

zhangh43:
## Tests

no test needed, just fix pipeline.


HaozhouWang:
As required by RelEng, we may need to change the filename to a versioned filename. Please hold this PR until we get updated info from RelEng.

PR #494, opened 2019-08-15
Fixed answer file when gp_inject_fault changes return value

gfphoenix78:
## Description

In the old version, gp_inject_fault() returns true if success.
Now gp_inject_fault() returns "Success:" if success.

## Tests

Fix answer file


gfphoenix78:
I apologize, this commit is not complete. There is something missing in the next commit 3f2796039c0b4d4fca1ce8711bd75886a7c90c34

PR #497, opened 2019-08-21
Refactor server side connection code

HaozhouWang:
1. Decouple network connection related functions into a separate file
2. Remove hardcode type define and use gnu99 type (same as GPDB side)
   in server-side (R only)
3. Separate misc functions into GPDB related and server related.
4. Unify the plcConn and only use the global variable.
5. Cleanup head file.
6. Remove useless pre-set env variables as the coordinator will store
   such information.

Co-authored-by: Violet Cheng <vcheng@pivotal.io>
Co-authored-by: Haozhou Wang <hawang@pivotal.io>


src/common/comm_misc.h
```diff
@@ -0,0 +1,46 @@
...
+ * Portions Copyright © 2016-Present Pivotal Software, Inc.
+ */
+
+#ifndef PLC_COMM_MISC_H
+#define PLC_COMM_MISC_H
+
+#include "postgres.h"
+
+
+#define plc_elog(lvl, fmt, ...) elog(lvl, "plcontainer: " fmt, ##__VA_ARGS__);
```
gfphoenix78: I think `plc_elog` could be a function rather than a macro. So in any source file, we only need to declare a function instead of `#include` a header file. And this may decrease the number of macro test.

src/common/comm_messages.c
```diff
@@ -32,7 +32,12 @@ interpreted as representing official policies, either expressed or implied, of t
 
 #include <stdlib.h>
 
-#include "comm_utils.h"
+#ifdef PLC_CLIENT
```
gfphoenix78: Can we figure out this conditional include lines in one place? Too many header files contain this ugly code.

src/common/comm_channel.c
```diff
@@ -43,12 +47,12 @@ interpreted as representing official policies, either expressed or implied, of t
...
-static int send_uint32(plcConn *conn, uint32 i);
-static int send_int64(plcConn *conn, int64 i);
-static int send_float4(plcConn *conn, float4 f);
-static int send_float8(plcConn *conn, float8 f);
+static int send_int16(plcConn *conn, int16_t i);
+static int send_int32(plcConn *conn, int32_t i);
+static int send_uint32(plcConn *conn, uint32_t i);
+static int send_int64(plcConn *conn, int64_t i);
+static int send_float4(plcConn *conn, float f);
+static int send_float8(plcConn *conn, double f);
```
gfphoenix78: Is it safe to be sure that `sizeof(float)==4 && sizeof(double)==8` ?  It's better to add an assert somewhere.

HaozhouWang: Thanks fixed.

src/common/comm_messages.c
HaozhouWang: We use a separate include folder to solve this issue.

src/common/comm_misc.h
HaozhouWang: Agree, but it seems out of scope in this PR. We can have a better solution in the next PR to total fix the issue of the log module.

src/include/plc/misc.h
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#ifndef PLC_COMM_MISC_H
+#define PLC_COMM_MISC_H
+
+#include "postgres.h"
+#include <unistd.h>
+#include <stdint.h>
+
+#define plc_elog(lvl, fmt, ...) elog(lvl, "plcontainer: " fmt, ##__VA_ARGS__);
```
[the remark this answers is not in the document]
HaozhouWang: I agree to extract as an independent function. We can do this in the next PR.

HaozhouWang:
Thanks for reviewing, pushed.

PR #500, opened 2019-08-26
Add initialization for coordinator

gfphoenix78:
## Description

Initialize coordinator, includes:
* init shared memory, (memory for storing inspecting data is not ready now)
* starts two background workers, one is for the coordinator and the other is for auxiliary process
* init UNIX domain socket for the coordinator to listen to messages from QE/plcontainer
* init loop framework

Co-authored-by: Xinyue Ge <xge@pivotal.io>
Co-authored-by: Hao Wu <hawu@pivotal.io>


src/co_coordinator.c
```diff
@@ -33,108 +33,165 @@
...
 
-void
-plc_coordinator_main(Datum datum)
+static int
+plc_listen_socket()
 {
-    // TODO: impl coordinator logic here
+    char address[500];
+    int sock;
+    snprintf(address, sizeof(address), "/tmp/plcoordinator_%d_unix.sock", (int)getpid());
```
HaozhouWang: It is better to change to .plcoordinator.%d.unix.sock and hidden from the user.

src/co_coordinator.c
```diff
@@ -33,108 +33,165 @@
...
+        /* TODO: add network code */
+
+        rc = WaitLatch(&MyProc->procLatch, WL_LATCH_SET | WL_TIMEOUT | WL_POSTMASTER_DEATH, 0);
+        if (rc & WL_POSTMASTER_DEATH)
+            break;
+        ResetLatch(&MyProc->procLatch);
+        if (got_sighup) {
+            plc_refresh_container_config(false);
+        }
+        sleep(2);
```
HaozhouWang: It is better to add a comment here why we sleep 2 seconds?

src/co_coordinator.h
```diff
@@ -0,0 +1,34 @@
+#ifndef _CO_COORDINATOR_H
```
HaozhouWang: Add license header

src/co_coordinator.h
```diff
@@ -0,0 +1,34 @@
...
+    coordinator_protocol protocol;
+    char address[504];
+} CoordinatorStruct;
+#define CO_SHM_KEY  "plcoordinator_shm"
+
+typedef struct
+{
+    int id; // pid of QE
+    int sock; // socket file descriptor between QE and coordinator
+} requester_info_entry;
```
HaozhouWang: Do we need to store the id-sock pair in memory? If yes, why? It seems that the connection will be closed after the coordinator sends results back to GPDB and will not be reused anymore.

HaozhouWang:
It is better to move coordinator code into a single folder.

src/co_coordinator.c
gfphoenix78: It's a dummy line here. To make the function not be a busy loop

src/co_coordinator.c
gfphoenix78: OK, thanks!

src/co_coordinator.h
gfphoenix78: Thank you!

src/co_coordinator.h
gfphoenix78: This struct entry in a WIP status. The connection may be or be not closed after the first allocation of container. Maybe there is some other information to add that can't figure out now. In the other hand, if the socket file descriptor is needless, the whole hash map can also be removed.
The next step will determine whether to keep or remove it.

HaozhouWang: Sounds good!

src/co_coordinator.c
HaozhouWang: Got it!

PR #505, opened 2019-09-05
plcontainer connection to uds and sending message

violet2016:
1. plcontainer read sock address from shm
2. plcontainer send a struct with sessionid, processid, runtimeid

receive message is not added yet.

### DIFF
--- a/src/containers.c
+++ b/src/containers.c
@@ -30,6 +30,7 @@
 #include "common/comm_channel.h"
 #include "common/comm_connectivity.h"
 #include "common/comm_dummy.h"
+#include "common/comm_shm.h"
 #include "common/messages/messages.h"
 #include "plc/plc_configuration.h"
 #include "plc/containers.h"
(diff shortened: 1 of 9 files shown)


src/containers.c
```diff
@@ -119,15 +120,28 @@ static int get_new_container_from_coordinator(const char *runtime_id, plcContext
 {
 	plcConn *conn; // a connection used to connect to coordinator
 	int ret = 0;
+	int res = 0;
 	(void)runtime_id;
```
HaozhouWang: this line can be removed now

src/containers.c
```diff
@@ -201,7 +215,11 @@ static int init_container_connection(plcContext *ctx)
 
 // TODO: read shm to get the address of coordinator
 static char *get_coordinator_address(void) {
-	return NULL;
+	bool found;
+	coordinator_shm = ShmemInitStruct(CO_SHM_KEY, MAXALIGN(sizeof(CoordinatorStruct)), &found);
+    Assert(found);
```
HaozhouWang: indent

PR #507, opened 2019-09-11
coordinator receive message from plcontainer

violet2016:
1. add uds network connection in coordinator
2. process id message from plcontainer

### DIFF
--- a/src/plc_coordinator.c
+++ b/src/plc_coordinator.c
@@ -42,6 +42,7 @@
 #include "plc/plc_configuration.h"
 #include "plc/plc_coordinator.h"
 #include "common/comm_shm.h"
+#include "common/messages/messages.h"
 
 PG_MODULE_MAGIC;
 // PROTOTYPE:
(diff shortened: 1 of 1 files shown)


src/plc_coordinator.c
```diff
@@ -136,9 +141,63 @@ plc_initialize_coordinator()
...
+    (void) ptr;
+    ssize_t sz = 0;
+    struct sockaddr_un remote;
+    size_t len_addr = sizeof(struct sockaddr_un);
+    int s2 = accept(sock, &remote, &len_addr);
+    if (s2 < 0) {
+        return;
+    }
+    if ((sz=recv(s2, &msg_type, 1, 0)) < 0)
+    {
```
yydzero: Shall we process some special error, such as EAGAIN, EINTR, which just means need to retry. GPDB has lots of such kind of code, could use as a reference.

src/plc_coordinator.c
```diff
@@ -136,9 +141,63 @@ plc_initialize_coordinator()
...
+    int rv;
+    fd_set fdset;
+
+    FD_ZERO(&fdset);    /* clear the set */
+    FD_SET(sock, &fdset); /* add our file descriptor to the set */
+    timeout.tv_sec = TIMEOUT_SEC;
+    timeout.tv_usec = 0;
+
+    rv = select(sock + 1, &fdset, NULL, NULL, &timeout);
+    if (rv == -1) {
```
yydzero: same, shall we process special error number such as EAGAIN, EINTR?

violet2016: Got it. We'll improve the code in next pr.

PR #510, opened 2019-09-24
coordinator can terminate the R server process

violet2016:
When error happened in gpdb, coordinator can terminate the correspond R server process

1. add action in plcontainer control message to coordinator
2. add destroy_container in coordinator
3. send destroy message to aux process
4. add hash table to store the container/process info
(in stand alone mode the hash table is in coordinator, otherwise it is in aux)
5. kill R server process in stand alone mode

Tests can be covered by e2e test.

### DIFF
--- a/src/plc_coordinator.c
+++ b/src/plc_coordinator.c
@@ -72,14 +72,15 @@ char *plcontainer_stand_alone_server_path;
 
 
 static int start_container(const char *runtimeid, pid_t qe_pid, int session_id, int ccnt, char **uds_address);
+static int destroy_container(pid_t qe_pid, int session_id);
 static int send_message(QeRequest *request);
 static int receive_message();
+static int handle_request(QeRequest *req);
 static int start_stand_alone_process(const char* uds_address);
 static void shm_message_queue_receiver_init(dsm_segment *seg);
 static dsm_handle shm_message_queue_sender_init();
 
-
-// HTAB *container_status_table;
+HTAB *container_status_table;
 
 static void
 plc_coordinator_shmem_startup(void)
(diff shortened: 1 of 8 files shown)


src/plc_coordinator.c
```diff
@@ -336,36 +366,104 @@ _PG_init(void)
...
 
+static int store_container_info(ContainerKey *key, pid_t server_pid, char* container_id)
+{
+	ContainerEntry *entry = NULL;
+	bool found = false;
+	entry = (ContainerEntry *) hash_search(container_status_table, key, HASH_ENTER,
+											&found);
+	if (!found || (found && entry->stand_alone_pid == 0)) {
+		entry->stand_alone_pid = server_pid;
+	} else {
```
HaozhouWang: I think we may only need to check whether `entry` is null or not. If NULL, we throw out an error. Otherwise, we could just overwrite the old value as this only serves for positive destroy.

src/plc_coordinator.c
```diff
@@ -336,36 +366,104 @@ _PG_init(void)
...
+static int clear_container_info(ContainerKey *key)
+{
+	ContainerEntry *entry = NULL;
+	bool found = false;
+	int res;
+	entry = (ContainerEntry *) hash_search(container_status_table, key, HASH_FIND,
+											&found);
+	if (found && entry->stand_alone_pid != 0) {
+		elog(LOG, "try to terminate server process %d", entry->stand_alone_pid);
+		res = kill(entry->stand_alone_pid, SIGTERM);
```
HaozhouWang: I think we can use force kill (SIGKILL) here.

src/plc_coordinator.c
violet2016: OK, I'll fix this in next story, start a container will also need change in this  part

src/plc_coordinator.c
violet2016: Thanks, fixed

HaozhouWang:
LGTM

PR #511, opened 2019-09-26
fix pipeline build issue

wengyanqing:
## Description
Fix build pipeline issue. Use getenv function to avoid dependency on specific plcontainer extension code. 

## Tests
Since it's a compile issue. Existing regression testing is already enough.

### DIFF
--- a/src/common/comm_server.c
+++ b/src/common/comm_server.c
@@ -199,7 +199,7 @@ int start_listener() {
 	if (strcasecmp("true", use_container_network) == 0) {
 		sock = start_listener_inet();
 	} else if (strcasecmp("false", use_container_network) == 0){
-		if ((CurrentBackendType == BACKEND_DOCKER) && (geteuid() != 0 || getuid() != 0)) {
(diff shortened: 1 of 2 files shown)


wengyanqing:
@HaozhouWang @gfphoenix78 please help review.

src/common/comm_server.c
```diff
@@ -199,7 +199,7 @@ int start_listener() {
 	if (strcasecmp("true", use_container_network) == 0) {
 		sock = start_listener_inet();
 	} else if (strcasecmp("false", use_container_network) == 0){
-		if ((CurrentBackendType == BACKEND_DOCKER) && (geteuid() != 0 || getuid() != 0)) {
```
gfphoenix78: Is there any problem to check `CurrentBackendType`? It's wired to add an additional env for this issue.

wengyanqing: CurrentBackendType is a global variable in GPDB side, could not refer in pyclient/rclient side, thus it happens a compile issue.

gfphoenix78: Got it. But it's better to set an env in every server process(docker mode or process mode).

PR #512, opened 2019-09-27
Start Container in non stand alone mode

violet2016:
1. coordinator call plc docker related api to start container
2. change docker api, remove unused username & db name
3. fix uds path in container mode

Tests are covered by e2e tests


HaozhouWang:
suggest removing plc_backend_api.c in this PR as well. As we do not have plan to support other container tools other than docker.

src/plc_coordinator.c
```diff
@@ -407,20 +407,34 @@ static int start_container(const char *runtimeid, pid_t qe_pid, int session_id,
...
 	int res;
 
-	(void) runtimeid;
 	*uds_address = (char*) palloc(1024);
-	snprintf(*uds_address, 1024, "%s.%d.%d.%d.%d", DEBUG_UDS_PREFIX, qe_pid, session_id, ccnt, (int)getpid());
 
 	/* debug test only, we need to store the container info in coordinator */
 	if (plcontainer_stand_alone_mode)
 	{
+		snprintf(*uds_address, 1024, "%s.%d.%d.%d.%d", DEBUG_UDS_PREFIX, qe_pid, session_id, ccnt, (int)getpid());
```
wengyanqing: It's better to use a macro to replace such magic number. Also includes some other places.

src/include/plc/plc_docker_api.h
```diff
@@ -26,7 +26,7 @@ typedef struct {
 	int status;
 } plcCurlBuffer;
 
-int plc_docker_create_container(runtimeConfEntry *conf, char **name, int container_slot, char **uds_dir);
+int plc_docker_create_container(runtimeConfEntry *conf, char **name, char **uds_dir, pid_t qe_pid, int session_id, int ccnt);
```
wengyanqing: It's better to use a struct for the create container parameter (qe_pid, session_id, ccnt).

PR #515, opened 2019-10-09
Update pipeline: replace aggregate by in_parallel

gfphoenix78:
## Description

Replace aggregate by in_parallel
Remove some useless aggregate
## Tests

Update pipeline yaml only, no feature to test.


beeender:
ytt is used for the pipeline template. not needed anymore.

PR #518, opened 2019-10-21
Track containers' status and delete exited containers

violet2016:
1. add container status hash map in aux process
2. add entry in hash map when a container is created
3. check container status, delete exited container

test is covered by e2e tests


src/plc_coordinator.c
```diff
@@ -547,6 +549,36 @@ static int receive_message()
...
+{
+    int entry_num = 0;
+    int res = 0;
+    int i = 0;
+    entry_num = hash_get_num_entries(container_status_table);
+    if (entry_num == 0) {
+        return 0;
+    }
+    HASH_SEQ_STATUS scan;
+	ContainerEntry *container_entry = NULL;
```
HaozhouWang: indent

PR #519, opened 2019-10-23
fix stability bugs

violet2016:
1. fix minor bugs that could cause crashes during stability tests
2. disable python test in stability tests

### DIFF
--- a/src/plc_configuration.c
+++ b/src/plc_configuration.c
@@ -650,10 +650,10 @@ char *get_sharing_options(runtimeConfEntry *conf, bool *has_error, char **uds_di
 			if (i > 0)
 				comma = ',';
 			if (conf->sharedDirs[i].mode == PLC_ACCESS_READONLY) {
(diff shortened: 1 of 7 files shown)


src/plc_configuration.c
```diff
@@ -673,8 +673,8 @@ char *get_sharing_options(runtimeConfEntry *conf, bool *has_error, char **uds_di
 
 			if (i > 0)
 				comma = ',';
-			
-			volumes[i] = palloc(10 + strlen(uds_dir) + strlen(IPC_CLIENT_DIR));
+
+			volumes[i] = palloc(10 + strlen(*uds_dir) + strlen(IPC_CLIENT_DIR));
 			sprintf(volumes[i], " %c\"%s:%s:rw\"", comma, *uds_dir, IPC_CLIENT_DIR);
```
HaozhouWang: Could we change to snprintf to avoid overhead

violet2016: fixed

PR #525, opened 2019-10-29
Introduce GRPC & Protobuf to replace communication framework

wengyanqing:
## Description
1. Introduce GRPC & Protobuf to replace legacy communication framework.
2. Implement scalar data type ser/des.
3. Disable init_container_connection for rserver and will be refactor later.

## Tests
Since it's just refactor communication framework, all the test cases are enough to cover.


src/containers.c
```diff
@@ -103,21 +103,23 @@ static plcContext *get_new_container_ctx(const char *runtime_id)
 	plcContextInit(ctx);
 
 	res = get_new_container_from_coordinator(runtime_id, ctx);
-
+    sleep(2);
```
wengyanqing: It just a workaround to remove init_container_connection because it needs to wait rserver process started.

PR #534, opened 2019-11-13
Add UDT support for plcontainer side.

wengyanqing:
## Description
1. plcontainer function parameter support UDT.
2. Remove plcUDT message type.
3. Fix standalone mode execl issue.

## Tests
Existing test cases are covered by UDT testing.

## Additional Notes
Since RServer has not supported UDT currently, the function return UDT will be enabled later.


wengyanqing:
> It is better to add some comments to each function.

Sure, will add comments later.

PR #570, opened 2020-02-04
refresh configuration whenever the config file is changed

violet2016:

### DIFF
--- a/src/plc_coordinator.c
+++ b/src/plc_coordinator.c
@@ -7,6 +7,7 @@
 
 #include "postgres.h"
 #include <unistd.h>
+#include <sys/inotify.h>
 
 #include "access/tupdesc.h"
 #include "access/xact.h"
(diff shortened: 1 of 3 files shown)


src/plc_coordinator.c
```diff
@@ -176,17 +177,82 @@ plc_initialize_coordinator(dsm_handle seg)
...
+
+		/* Read some events. */
+
+		len = read(fd, buf, sizeof(buf));
+		if (len == -1 && errno != EAGAIN) {
+			elog(ERROR, "failed to read config file.");
+			return;
+		}
+
+		/* If the nonblocking read() found no events to read, then
```
HaozhouWang: is this comment belonging to upper codes?

src/plc_coordinator.c
```diff
@@ -176,17 +177,82 @@ plc_initialize_coordinator(dsm_handle seg)
...
+	ssize_t len;
+	char *ptr;
+
+	/* Loop while events can be read from inotify file descriptor. */
+
+	for (;;) {
+
+		/* Read some events. */
+
+		len = read(fd, buf, sizeof(buf));
```
HaozhouWang: if the read is non-blocking, I think we may add a pg_sleep (e.g every second) here for the frequent checking.

src/plc_coordinator.c
violet2016: The comment is from inotify sample code.

PR #596, opened 2020-04-03
Add new option --use_local_copy in utils

HaozhouWang:
1. add a new option --use_local_copy(-ulc) to allow docker load the
   image file from local copy. With this option, plcontainer utils
   will not try to distribute the docker image file from master node
   to all segment/standby-master nodes.
2. improve the error message for runtime-add
3. update the concourse test


mkiyama:
@HaozhouWang 
Hi, could you explain why I would use this option?

thanks

pivotalguru:
@mkiyama Deployments in the cloud use a single image to create the VMs for the cluster.  Inside the image will be all of the binaries including the docker images.  In this environment, there is no need to copy the docker image to all of the deployed VMs because each VM already has it.  We just need to register the docker images on the nodes and skip copying it there.

mkiyama:
Thanks, Jon

PR #602, opened 2020-05-08
remove coordinator for pl4k mode

violet2016:
## Description

1. build plcontainer with "make pl4k PL4K=yes" and install it with "make installpl4k PL4K=yes"
2. please add "plcontaienr.so" to greenplum preload libraries

### DIFF
--- a/src/plc_configuration.c
+++ b/src/plc_configuration.c
@@ -703,7 +703,12 @@ containers_summary(pg_attribute_unused() PG_FUNCTION_ARGS) {
 			actualLen++;
 		}
 		funcctx->max_calls = (uint32_t) actualLen;
+#ifndef PL4K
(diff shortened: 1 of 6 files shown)


src/plc_configuration.c
```diff
@@ -703,7 +703,11 @@ containers_summary(pg_attribute_unused() PG_FUNCTION_ARGS) {
 			actualLen++;
 		}
 		funcctx->max_calls = (uint32_t) actualLen;
+#ifndef PL4K
 		res = PlcDocker_stat(ids, actualLen, mem_usage);
+#else
+		res = -1;
+#endif
```
HaozhouWang, line 710: Could we change res to -2 with adding a new message to the user say like un-support feature in K8S environment?

PR #612, opened 2020-05-15
Dockerfile.python3.ubuntu: reduce size

yvaa:
Hello!
1. Switched image to `python:3.7-slim-buster` - this image is smaller than `ubuntu:18.04` and python 3.7 already installed
2. Removed `python3.7-dev` installation. It looks like this package came from developing process and not needed to run pyclient.

As result the image size will be 2-3x smaller.

Please use https://hub.docker.com/_/python/ images in the future. For example there is `python:3.7-alpine` image - it's extremely small image. But it requires to build packages during install using `pip` and more fine-tunning the image so I choose the `slim-buster` for this pull request.

### DIFF
--- a/dockerfiles/Dockerfile.python3.ubuntu
+++ b/dockerfiles/Dockerfile.python3.ubuntu
@@ -1,7 +1,6 @@
-FROM ubuntu:18.04
+FROM python:3.7-slim-buster
 
 RUN mkdir -p /clientdir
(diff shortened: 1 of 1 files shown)


HaozhouWang:
@yvaa Thanks for your PR. After the team discussion, we plan to keep the ubuntu image as it provides more flexibility for further extension.

PR #615, opened 2020-10-23
Add new UDF and views to collect containers information

HaozhouWang:
1. Add new UDF plcontainer_containers_info to collect container information
2. Update a view to show information directly.


violet2016 (approved):
+1

PR #617, opened 2020-11-13
Information collector now support to show currently running UDF

HaozhouWang:
1. Information collector can show currently the name of UDF, which is running in the container
2. The plcontainer must be loaded in shared_preload_library GUC
3. Guc ` plcontainer.enable_function_collection` must set to TURE to enable the UDF collector


violet2016 (approved):
+1

lenapivotal:
@HaozhouWang is it possible to give me access to your dev tracker so we can mark this story as a feature that needs docs? And we can add to the gpdb docs tracker?

HaozhouWang:
@lenapivotal 
Sure, the invitation is sent.

PR #622, opened 2021-11-01
Fix memory related errors

RMTT:
## Description
Fix memory-related errors when using sql to select array type columns or large size data.

## Tests
Create a table with an array type column, and select it in plcontainer

## User Story or Issue
https://www.pivotaltracker.com/story/show/179879954


src/sqlhandler.c
```diff
@@ -63,7 +63,14 @@ static plcMsgResult *create_sql_result(bool isSelect) {
 	result->types = palloc(result->cols * sizeof(*result->types));
 	result->names = palloc(result->cols * sizeof(*result->names));
 	result->exception_callback = NULL;
+
+        /*
```
beeender, line 67: indention is wrong

src/sqlhandler.c
```diff
@@ -310,6 +312,11 @@ plcMessage *handle_sql_message(plcMsgSQL *msg, plcConn *conn, plcProcInfo *pinfo
 						}
 					}
 
+                                        /*
```
beeender, line 315: indention is wrong

src/sqlhandler.c
```diff
@@ -63,7 +63,14 @@ static plcMsgResult *create_sql_result(bool isSelect) {
...
+        /*
+         * We do not need free this array because the array iterator will reference it.
+         * This function is under SPI_proc Memory Context, when SPI_finish is invoked, 
+         * the array will be freed automatically. 
+         */
 	resTypes = palloc(result->cols * sizeof(plcTypeInfo));
+
 	for (j = 0; j < result->cols; j++) {
 		fill_type_info(NULL, SPI_tuptable->tupdesc->attrs[j]->atttypid, &resTypes[j]);
 		copy_type_info(&result->types[j], &resTypes[j]);
```
beeender, line 76: If the type is not going to be freed, do we still need to copy them here?

xuebinsu (approved):
+1, better to keep indention consistent.

src/sqlhandler.c
```diff
@@ -63,7 +63,14 @@ static plcMsgResult *create_sql_result(bool isSelect) {
 	result->types = palloc(result->cols * sizeof(*result->types));
 	result->names = palloc(result->cols * sizeof(*result->names));
 	result->exception_callback = NULL;
+
+        /*
+         * We do not need free this array because the array iterator will reference it.
+         * This function is under SPI_proc Memory Context, when SPI_finish is invoked, 
+         * the array will be freed automatically. 
+         */
```
beeender, line 71: I am thinking of another possible fix.
The `plcMsgResult` is supposed to be freed at `plcontainer.c:565` the `free_result()`.
Would it be better if we has `plcTypeInfo` as a member in the `plcMsgResult` and free it together in the `free_result()`? Then we have a completed life-control of the type. What do you think?

beeender: One more thing, do we have a test case to cover the array types here?

src/sqlhandler.c
RMTT: Yeah, because the `result->types` has a different structure and layout from `resTypes`.

beeender: Would it make sense if we just have `plcTypeInfo` in the `plcMsgResult`

```
typedef struct plcMsgResult {
...
	// Remove this
	// plcType *types;
	// Add
	plcTypeInfo *types;
...
} plcMsgResult;
```

So we don't have to copy, and the types can be freed in `free_result()`?

src/sqlhandler.c
beeender: @RMTT Just install this one
https://github.com/editorconfig/editorconfig-vim

beeender:
The minimal reproduce steps for the original issue:

```sql
DROP TABLE IF EXISTS array_test_table;
CREATE TABLE array_test_table(
    content text[]
)
DISTRIBUTED RANDOMLY;
INSERT INTO array_test_table(content)
VALUES('{8.03,8.033,8.305,8.336,0.838,0.883,8.53,8.586,0.880,0.833,0.883,0.588,0.588,3.885,3.3,3.338,063.858,053.008,80.336,LP,8.655,8.858,8.803,8.608,00.385,88.865,85.088,88.603,85.300,58.086}');

CREATE OR REPLACE FUNCTION array_test()
RETURNS text
VOLATILE
AS $$
# container: plc_python3_shared
import pandas as pd
with plpy.subtransaction():
    df = pd.DataFrame.from_records(
    plpy.execute(
    """
     select * from array_test_table;
    """))
    return "OK"
$$
LANGUAGE plcontainer
EXECUTE ON ANY;
select * from array_test();

```

src/sqlhandler.c
```diff
@@ -310,6 +312,11 @@ plcMessage *handle_sql_message(plcMsgSQL *msg, plcConn *conn, plcProcInfo *pinfo
 						}
 					}
 
+                    /*
```
beeender, line 315: indention

tests/Makefile
```diff
@@ -34,7 +34,7 @@ resgroup:
 	plcontainer runtime-show -r plc_r_shared
 
 	PL_TESTDB=$(PL_TESTDB) $(top_builddir)/src/test/regress/pg_regress \
-				$(PSQLDIR) $(RESGROUP_OPTS) || if [[ -f regression.diffs ]]; then cat regression.diffs; exit 1; fi
+				$(PSQLDIR) $(RESGROUP_OPTS)
```
beeender, line 37:
```
PostgreSQL regression test driver

Usage:
  pg_regress [OPTION]... [EXTRA-TEST]...

Options:
...
  --print-failure-diffs     Print the diff file to standard out after a failure
```

There is an option for this. Try it

src/sqlhandler.c
```diff
@@ -63,7 +63,14 @@ static plcMsgResult *create_sql_result(bool isSelect) {
 	result->types = palloc(result->cols * sizeof(*result->types));
 	result->names = palloc(result->cols * sizeof(*result->names));
 	result->exception_callback = NULL;
+
+    /*
+    * We do not need free this array because the array iterator will reference it.
```
beeender, line 68: indention

tests/Makefile
beeender: this is good. Ignore me

beeender (approved):
:clap:

PR #626, opened 2022-03-24
CI: add mirrorless pipeline for story

yihong0618:
This PR add mirrorless  pipeline for story #[181397602](https://www.pivotaltracker.com/story/show/181397602)

backport sql_isolation_testcase.py from gpdb6 to plcontainer to fix ci down --> this also a bug for 6X stable

## Note:
need change deafault branch for pipeline when merge change it to 6X_STABLE

### DIFF
--- a/tests/isolation2/sql_isolation_testcase.py
+++ b/tests/isolation2/sql_isolation_testcase.py
@@ -16,50 +16,280 @@ limitations under the License.
 """
 
 import pygresql.pg
+import pty
 import os
-import subprocess
+try:
+    import subprocess32 as subprocess
+except:
+    import subprocess
 import re
 import multiprocessing
+import tempfile
 import time
 import sys
 import socket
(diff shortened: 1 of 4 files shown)


tests/isolation2/sql_isolation_testcase.py
```diff
@@ -16,50 +16,280 @@
 """
```
Sasasu, line 16: I think this is good enough. In the future may delete this file, but not now.

Sasasu (approved):

beeender (approved):
Thanks for this!

And do we have a testing pipeline somewhere? cannot find it.

yihong0618:
> Thanks for this!
> 
> And do we have a testing pipeline somewhere? cannot find it.

here:
https://extensions.ci.gpdb.pivotal.io/teams/main/pipelines/hy_mirror_less_fix

PR #627, opened 2022-04-28
fix compile error

Sasasu:
compile error when using clang 13
```
src/plc_container_info.c:713:19: error: variable 'per_query_ctx' set but not used [-Werror,-Wunused-but-set-variable]
    MemoryContext       per_query_ctx;
```

```
src/plcontainer.c:176:8: error: variable 'is_trigger' set but not used [-Werror,-Wunused-but-set-variable]
        bool            is_trigger;
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 #------------------------------------------------------------------------------
-# 
+#
 # Copyright (c) 2016-Present Pivotal Software, Inc
 #
 #------------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -104,24 +104,22 @@ else
   override CLIENT_CFLAGS += -O3 -g
 endif
 
-# detected the docker API version, only for centos 6
-RHEL_MAJOR_OS=$(shell cat /etc/redhat-release | sed s/.*release\ // | sed s/\ .*// | awk -F '.' '{print $$1}' )
-ifeq ($(RHEL_MAJOR_OS), 6)
-  override CFLAGS +=  -DDOCKER_API_LOW
```
Sasasu, line 110: this macro is never used.
and the meaning for `DOCKER_API_LOW` is confused.

PR #628, opened 2022-04-28
Enable network in container

Sasasu:

### DIFF
--- a/management/bin/plcontainer
+++ b/management/bin/plcontainer
@@ -593,6 +593,11 @@ def validate_xml(filename):
                     if use_container_logging_str != 'yes' and use_container_logging_str != 'no':
                         logger.error("'use_container_logging' should be 'yes' or 'no' in runtime %s, but now: '%s'", runtime_id, use_container_logging_str)
                         raise Exception("Validation failed")
+                elif 'enable_network' in settings.attrib:
+                    enable_network_str = settings.attrib['enable_network'].lower()
+                    if enable_network_str != 'yes' and enable_network_str != 'no':
+                        logger.error("'enable_network' should be 'yes' or 'no' in runtime %s, but now: '%s'", runtime_id, enable_network_str)
+                        raise Exception("Validation failed")
                 elif 'resource_group_id' in settings.attrib:
                     resource_group_id_str = settings.attrib['resource_group_id']
                     if resource_group_id_str.isdigit() != True:
@@ -760,7 +765,7 @@ def get_conf_settings(options, elements):
         strList = setting.split("=")
         if len(strList) != 2:
             raise Exception("Bad setting format: %s" % setting)
-        if strList[0] != "memory_mb" and strList[0] != "cpu_share" and strList[0] != "use_container_logging" and strList[0] != "resource_group_id" and strList[0] != "roles":
+        if strList[0] not in ("memory_mb", "cpu_share", "use_container_logging", "resource_group_id", "roles", "enable_network"):
             raise Exception("Bad setting key: %s" % strList[0])
         elements['setting'][strList[0]] = strList[1]
 
@@ -944,7 +949,7 @@ def main():
 
         if options.subparser_name == "runtime-edit":
             if options.editor is None:
-                options.editor  = os.getenv("EDITOR") or "vi" # default editer is 'vi'
+                options.editor  = os.getenv("EDITOR", "vi") # default editer is 'vi'
             changed = edit_file(tmp_file, options.editor)
             if changed:
                 logger.debug('Preparing the worker pool...')
--- a/management/bin/plcontainer
+++ b/management/bin/plcontainer
@@ -944,7 +944,7 @@ def main():
 
         if options.subparser_name == "runtime-edit":
             if options.editor is None:
-                options.editor  = "vi"
+                options.editor  = os.getenv("EDITOR") or "vi" # default editer is 'vi'
             changed = edit_file(tmp_file, options.editor)
             if changed:
                 logger.debug('Preparing the worker pool...')
--- a/management/config/plcontainer_configuration.xml
+++ b/management/config/plcontainer_configuration.xml
@@ -24,6 +24,8 @@
                  When not set, the default CPU share is 1024.
             6.3. "use_container_logging" - set to "yes" or "no" for container logging (not for backend)
                  By default, we set "no".
+            6.4. "enable_network" - set to "yes" or "no", default value is "no".
+                 when "enable_network" is "no", the inside UDF did not have network access.
         All the container images not manually defined in this file will not be
         available for use by endusers in PL/Container
     -->
--- a/src/plc_configuration.c
+++ b/src/plc_configuration.c
@@ -174,6 +174,7 @@ static void parse_runtime_configuration(xmlNode *node) {
 		conf_entry->cpuShare = 1024;
 		conf_entry->useContainerLogging = false;
 		conf_entry->useContainerNetwork = false;
+		conf_entry->enableNetwork = false;
 		conf_entry->resgroupOid = InvalidOid;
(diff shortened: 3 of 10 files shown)


Sasasu:
CI https://extensions.ci.gpdb.pivotal.io/teams/main/pipelines/DEV:SASASU:plcontainer-gpdb-6X-STABLE

management/config/plcontainer_configuration.xml
```diff
@@ -24,6 +24,8 @@
                  When not set, the default CPU share is 1024.
             6.3. "use_container_logging" - set to "yes" or "no" for container logging (not for backend)
                  By default, we set "no".
+            6.4. "enable_network" - set to "yes" or "no", default value is "no".
+                 when "enable_network" is "no", the inside UDF did not have internet access.
```
beeender, line 28: `internet access.` not only the internet I think? intranet won't be available as well. So just `network access`?

src/plc_configuration.c
```diff
@@ -253,10 +254,37 @@ static void parse_runtime_configuration(xmlNode *node) {
...
+					value = xmlGetProp(cur_node, (const xmlChar *) "enable_network");
+					if (value != NULL) {
+						validSetting = true;
+						if (strcasecmp((char *) value, "yes") == 0) {
+							conf_entry->enableNetwork = true;
+						} else if (strcasecmp((char *) value, "no") == 0) {
+							conf_entry->enableNetwork = false;
+						} else {
+							plc_elog(ERROR, "SETTING element <enable_network> only accepted \"yes\" or"
+								"\"no\" only, current string is %s", value);
```
beeender, line 273: missing space between "or" and "no"
and quotes the `%s`, like `\"yes\"`

src/plc_configuration.c
```diff
@@ -253,10 +254,37 @@ static void parse_runtime_configuration(xmlNode *node) {
...
-					 * this should be set by various backend implementation.
-					 */
+
+					// Enforce to not use network for connection. In the future
+					// this should be set by various backend implementation.
 					conf_entry->useContainerNetwork = false;
+
+					// enable or disable network in container. default is no
+					// will be force yes when use_container_network = yes
+					value = xmlGetProp(cur_node, (const xmlChar *) "enable_network");
```
beeender, line 264: This whole function doesn't look neat at all, so many duplicated lines :(
But not the scope for this PR. Just some random complains.

src/plc_configuration.c
```diff
@@ -253,10 +254,37 @@ static void parse_runtime_configuration(xmlNode *node) {
...
+						} else {
+							plc_elog(ERROR, "SETTING element <enable_network> only accepted \"yes\" or"
+								"\"no\" only, current string is %s", value);
+						}
+
+						xmlFree((void *) value);
+						value = NULL;
+
+						// force enable when use_container_network = true
+						if (conf_entry->enableNetwork == false && conf_entry->useContainerNetwork == true) {
```
beeender, line 280: is this for the future?? `conf_entry->useContainerNetwork = false;` at line 260

src/plc_configuration.c
```diff
@@ -456,14 +484,16 @@ static void print_runtime_configurations() {
...
 			plc_elog(INFO, "    image = '%s'", conf_entry->image);
+			plc_elog(INFO, "    command = '%s'", conf_entry->command);
 			plc_elog(INFO, "    memory_mb = '%d'", conf_entry->memoryMb);
 			plc_elog(INFO, "    cpu_share = '%d'", conf_entry->cpuShare);
-			plc_elog(INFO, "    use container logging  = '%s'", conf_entry->useContainerLogging ? "yes" : "no");
-			if (conf_entry->useUserControl){
+			// skip conf_entry->useContainerNetwork because it is not user settable.
+			plc_elog(INFO, "    use_container_logging  = '%s'", conf_entry->useContainerLogging ? "yes" : "no");
+			plc_elog(INFO, "    enable_network  = '%s'", conf_entry->enableNetwork ? "yes" : "no");
+			if (conf_entry->useUserControl && conf_entry->roles != NULL){
```
beeender, line 493: space between `NULL)` and `{`

beeender (commented):
The code looks good in general. We may need to consider the safe and unsafe things which we have discussed. But that could be implemented in another PR.

Also @RMTT needs to review this.

management/bin/plcontainer
```diff
@@ -944,7 +949,7 @@ def main():
 
         if options.subparser_name == "runtime-edit":
             if options.editor is None:
-                options.editor  = "vi"
+                options.editor  = os.getenv("EDITOR") or "vi" # default editer is 'vi'
```
yihong0618, line 952: maybe work well too?
```
os.getenv("EDITOR", "vi")
```

yihong0618 (approved):

src/plc_configuration.c
yihong0618: > L

maybe we can bring `diskquota` fmt actions and logic to this repo?

Sasasu: I can't live without clang-format. 😱
I am very careful not to press the 'Format' button.

Sasasu:
for the trusted or untrusted PL problem:

in the very beginning of plcontainer. the network is enabled on purpose.
https://github.com/greenplum-db/plcontainer/blob/22aafb803d62de526ce1aba37b68b60e6f6b87c1/src/plc_docker_api.c#L40-L54

and plcontainer is a trusted PL.

network disabled after https://github.com/greenplum-db/plcontainer/commit/37b2b8875e9cd19f83f73c694d6f9745598af64f

this commit removed the `\"DisableNetwork\": false,\n"`, add `\"NetworkDisabled\": %s,\n"` and setting networkdisabled to true.
the commit wants to support unix domain socket. not related to the network. the disabled whole network looks like a bug.

and some doc copy from pljava

```

In the 1.6.0 default policy, 'trusted' (`java`) code has minimal permissions,
suitable for general computation and interacting with the database, and
'untrusted' (`javaU`) code has only the additional permission to access the
file system. Existing user functions that worked in PL/Java 1.5.x and performed
other actions, such as making network connections, will need the appropriate
permissions (such as `java.net.URLPermission` or `java.net.SocketPermission`)
granted via the policy file.
```

after pljava 1.6 (released on Oct 19, 2020). the trusted also can access network by adding `java.net.SocketPermission`.
so can not access the network is not a feature.

src/plc_configuration.c
Sasasu: how about the JSON api like
https://github.com/greenplum-db/plcontainer/blob/2f2dbd572efd0c84aed51da3ee44ad31328459d8/src/plc_docker_api.c#L217-L274

refactor it later!

src/plc_configuration.c
Sasasu: yes, for the future.

src/plc_configuration.c
RMTT: why not use `json-c``

RMTT (approved):
LGTM

xuebinsu:
It would be great if we could have some test cases that try to establish network connections.

Sasasu:
> > It would be great if we could have some test cases that try to establish network connections.
> 
> That is a good point.
> Is there any network API we can use in the container python? `httplib`? @Sasasu 

I want just do something like `os.exec("/bin/ping -n 1 127.0.0.1").read()`

i think this feature is not need a rush release, i will merge this pr after adding some basic testcase (and remove the QQQ debug log)

src/plc_configuration.c
Sasasu: json-c needs alloc memory and is not simple when serializing object array.
still need something like `__JSON_APPEND_WITH_NAME`.

I think the final json-c code will look similar to this code.

PR #632, opened 2022-06-10
4X performance impove

Sasasu:


yihong0618:
amazing....
Regex is that slow?

Sasasu:
context: plcontainer will compile this regex for every tuple. in PostgreSQL tuple process should fast than 1us.

regcomp will compile the regex, and do some JIT to generate machine code, regex_t has state so we can not reuse it.



when others do perf, they ignore anything not from PostgreSQL or plcontainer, I don't know why.

beeender (approved):
I think we can cache the `parse_container_meta` result so that there will be no need to call this multiple times in the same session. But it is beyond the scope of this PR.

This on is great, easy and free performance boost.

PR #633, opened 2022-06-10
set TCP_NODELAY=yes to speed our use case

Sasasu:
flush ASAP after any TCP send.
plcontainer has a buffer over TCP layer when server sends a message to client, we want client get this message immediately.

-----

localhost: 18.934s->19.238s (+1.5%)
not test on remotehost yet, but I think it will work.

### DIFF
--- a/src/common/comm_connectivity.c
+++ b/src/common/comm_connectivity.c
@@ -19,6 +19,7 @@
 #include <assert.h>
 #include <sys/un.h>
 #include <sys/socket.h>
+#include <netinet/tcp.h>
 #include <libgen.h>
 
 #include "comm_utils.h"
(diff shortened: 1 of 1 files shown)


Sasasu:
> Do we have a benchmark on bigger tuples?

the best test environment is a small TCP package with high latency.
this pr does not improve the performance for bigger tuples.

beeender:
> this pr does not improve the performance for bigger tuples.

I mean, will this make performance worse for bigger tuples?

Sasasu:
> I mean, will this make performance worse for bigger tuples?

no, the worst case is small tuples with `lo` interface. +1.5% is the worst case

beeender (approved):

PR #634, opened 2022-06-23
Replace rclient submodule with subtree

beeender:
It makes little sense to host the rclient code in a separate repo.
A license statement has been added to the README to indicate that the
rclient source has a different license (GPL) than others.
rclient source will be added as a full history git subtree later.


Sasasu (approved):

beeender:
Let me confirm with more people. will merge this later.

PR #635, opened 2022-06-23
cmake and use ytt to generate pipeline init

yihong0618:
This `PR` did these things:
1. use `ytt` to generate pipeline
2. open privilege to use docker in pipeline
3. add cmake now use cmake to `make pyclint`


CMakeLists.txt
```diff
@@ -0,0 +1,191 @@
...
+      CACHE PATH "default install prefix" FORCE)
+endif()
+
+set(DOWNLOAD_DIR
+    ${CMAKE_CURRENT_SOURCE_DIR}/downloads
+    CACHE STRING "Directory for downloading external project")
+
+# error on pipeline https://stackoverflow.com/questions/53633705/cmake-the-c-compiler-is-not-able-to-compile-a-simple-test-program
+SET (CMAKE_C_COMPILER_WORKS 1)
+SET (CMAKE_CXX_COMPILER_WORKS 1)
```
beeender, line 37: these two should not exist, they are not helping

CMakeLists.txt
```diff
@@ -0,0 +1,191 @@
...
+    INSTALL_DIR
+    ${deps_INSTALL_DIR}
+    CMAKE_ARGS
+    -D BUILD_SHARED_LIBS=OFF
+    -D BUILD_STATIC_LIBS=ON
+    -D CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+    -D CMAKE_INSTALL_PREFIX=${deps_INSTALL_DIR}
+    -D DISABLE_EXTRA_LIBS=ON
+    -D CMAKE_INSTALL_LIBDIR=lib
+    )
```
beeender, line 81: minor issue, `)` style doesn't match with others.

CMakeLists.txt
```diff
@@ -0,0 +1,191 @@
...
+    ${deps_INSTALL_DIR}/include
+    ${deps_INSTALL_DIR}/include/libxml2)
+target_link_libraries(plcontainer PRIVATE
+    ${deps_INSTALL_DIR}/lib/libxml2.a
+    ${deps_INSTALL_DIR}/lib/libjson-c.a)
+add_dependencies(
+     plcontainer
+     libxml
+     json-c
+)
```
beeender, line 103: ditto
and the indents are wrong

CMakeLists.txt
```diff
@@ -0,0 +1,191 @@
...
+
+# Build client
+set(PYCLIENT_IMAGE python39.alpine)
```
beeender, line 106: We needs this to be a cached cmake var so it can be changed when running cmake. But we can do that in a different PR.

CMakeLists.txt
```diff
@@ -0,0 +1,191 @@
...
+    PYTHONPATH=
+    docker run
+    --entrypoint=
+    -v ${CMAKE_CURRENT_BINARY_DIR}/rclient:/build
+    -v ${CMAKE_CURRENT_SOURCE_DIR}:/src
+    -w /build
+    -u ${my_UID}
+    -it ${RCLIENT_IMAGE}
+    sh -c 'cmake /src/src/rclient && cmake --build . -v')
+#add_dependencies(rclient rclient_image)
```
beeender, line 152: just enable this. I will push changes to the rclient repo soon.

dockerfiles/Dockerfile.python39.alpine
```diff
@@ -0,0 +1,9 @@
+FROM python:3.9-alpine3.13
+# FROM python:3.9-alpine
+# Why we choose alpine3.13 rather than alpine
+# because https://gitlab.alpinelinux.org/alpine/aports/-/issues/12321
```
beeender, line 4: It is better to put a general description of the problem first, then point to a link. Since the link could be dead someday, and we still need to know the reason when that happens.

Change to:

```
# apline >=3.14 has a problem with cmake which will report an error:
#     make: /usr/bin/make: Operation not permitted
# That happens with the docker on CentOS7.
# See more details on https://gitlab.alpinelinux.org/alpine/aports/-/issues/12321
```

concourse/scripts/build_plcontainer_cmake.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash -l
+# TODO when PR merge replace it with build_plcontainer.sh
+set -exo pipefail
+
+function build_rclient() {
+    [ -f /opt/gcc_env.sh ] && source /opt/gcc_env.sh
+    pushd /home/gpadmin/plcontainer_src
+    # init submoudle
+    git submodule update --init
```
beeender, line 9: this is not needed. concourse git resource should handle the submodule checkout.

concourse/scripts/build_plcontainer_cmake.sh
```diff
@@ -0,0 +1,24 @@
...
+    popd 
+
+    pushd /home/gpadmin/plcontainer_artifacts
+
+    cmake /home/gpadmin/plcontainer_src
+    cmake --build .
+    cmake --build . --target pyclient
```
beeender, line 16: we probably should move the cliient build to the test script to avoid build the image twice.
But not a problem for this PR.

concourse/scripts/docker-lib.sh
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+# Based on https://github.com/concourse/docker-image-resource/blob/master/assets/common.sh
```
beeender, line 2: put the source of the script here.
https://github.com/meAmidos/dcind

concourse/fly.sh
```diff
@@ -0,0 +1,128 @@
...
+      config_file="pr.yml"
+      hook_res="${proj_name}_pr"
+    ;;
+  commit)
+      if [ -z "${pipeline_name}" ]; then
+          pipeline_name="COMMIT:${proj_name}:main"
+      fi
+      # Default branch
+      if [ -z "${branch}" ]; then
+          branch="main"
```
beeender, line 65: wrong branch. it is still `6X_STABLE`

concourse/fly.sh
```diff
@@ -0,0 +1,128 @@
...
+      if [ -z "${pipeline_name}" ]; then
+          usage "'-p' needs to be supplied to specify the pipeline name for flying a 'dev' pipeline."
+      fi
+      pipeline_name="DEV:${pipeline_name}"
+      config_file="dev.yml"
+    ;;
+  release)
+      # Default branch is 'gpdb' as it is our main branch
+      if [ -z "${branch}" ]; then
+          branch="gpdb"
```
beeender, line 80: ditto

CMakeLists.txt
beeender: hmm, i changed mind, there are more work for r needs to be done. Just keep this disabled for now.  as well as rclient related install target.

Let's deal with rclient later.

beeender (approved):

PR #639, opened 2022-07-04
pyclient: support py object to UDT in table function.

Sasasu:
plpython support both Python Mapping (dictionary) and Python Object
to PostgreSQL UDT.

but plcontainer not support Python Object. add the support so
plcontainer can run the example code from PostgreSQL document.

https://www.postgresql.org/docs/14/plpython-data.html#id-1.8.11.11.6

create or replace function PyUDT() returns table(a float, b float)
as $$
    # container: python39.alpine:latest
    from collections import namedtuple
    A = namedtuple('A', 'a b')
    return [A(1,2)]
$$ language plcontainer;

# TODO

add basic test for this.


src/pyclient/pyconversions.c
```diff
@@ -491,38 +491,35 @@ static int plc_pyobject_as_array(PyObject *input, char **output, plcPyType *type
...
-			} else if (value == Py_None) {
-				udt->data[i].isnull = true;
-				udt->data[i].value = NULL;
-			} else {
-				udt->data[i].isnull = false;
-				res = type->subTypes[i].conv.outputfunc(value, &udt->data[i].value, &type->subTypes[i]);
-			}
-		}
+		py_get = PyObject_GetAttrString;
+	}
```
beeender, line 502:
```
Sequence type (a tuple or list, but not a set because it is not indexable)

    Returned sequence objects must have the same number of items as the composite result type has fields. The item with index 0 is assigned to the first field of the composite type, 1 to the second and so on. For example:

    CREATE FUNCTION make_pair (name text, value integer)
      RETURNS named_value
    AS $$
      return ( name, value )
      # or alternatively, as list: return [ name, value ]
    $$ LANGUAGE plpythonu;

    To return an SQL null for any column, insert None at the corresponding position.

    When an array of composite types is returned, it cannot be returned as a list, because it is ambiguous whether the Python list represents a composite type, or another array dimension.
```

Don't we want to support sequences as well?

src/pyclient/pyconversions.c
```diff
@@ -491,38 +491,35 @@ static int plc_pyobject_as_array(PyObject *input, char **output, plcPyType *type
...
-		*output = (char *) udt;
+	plcUDT *udt = pmalloc(sizeof(plcUDT));
+	udt->data = pmalloc(type->nSubTypes * sizeof(rawdata));
+	for (int i = 0; i < type->nSubTypes && res == 0; i++) {
+		PyObject *value = py_get(input, type->subTypes[i].typeName);
+		if (value == NULL) {
+			udt->data[i].isnull = true;
+			udt->data[i].value = NULL;
+			raise_execution_error("Cannot find key or attribute '%s' in result", type->subTypes[i].typeName);
+			res = -1;
```
beeender, line 512: shall we `break` here? the python exception has been thrown already, no need to continue the loop I think? And also `pfree` can be called here.

beeender (commented):
Nice catch!!

We need tests before merging this.

src/pyclient/pyconversions.c
Sasasu: > shall we break here?
no, this is py client code, it is a normal process without PostgreSQL runtime. `raise_execution_error` does not affect control flow.

> also pfree can be called here.
will add, that this is a memory leak.

Sasasu: can not add `pfree()` here. the flowing code will free memory.

the alloc stack:
```
> PLy_spi_execute_plan
> Ply_get_output_function
> arguments_to_pytuple
> plc_pyobject_as_udt / alloc(udt) + alloc(udt->data)
```

the free stack
```
> PLy_spi_execute_plan
> free_arguments
> plc_free_udt / pfree(udt->data[i].value)  + pfree(udt->data)
> free_result / pfree(udt);
```

tests/regress/expected/test_python3.out
```diff
@@ -653,7 +653,7 @@ select * from pytestudt13( (1,2,'a')::test_type3 );
 
 select pytestudt16();
 ERROR:  PL/Container client exception occurred: 
- Only 'dict' object can be converted to UDT "test_type3"
+ Exception occurred transforming result object to int8
```
Sasasu, line 656: `{'a': [1,3], 'b': [2,4], 'c': ['foo','bar']}` into `CREATE TYPE test_type3 AS (a int8, b float8, c varchar);` 

the error because `[1, 3]` into `a int8`

PR #641, opened 2022-07-08
feat: support ubuntu pipeline

yihong0618:
1. support ubuntu18
2. support rhe8

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,8 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Gpdb.cmake)
 include(${CMAKE_ROOT}/Modules/ExternalProject.cmake)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Git.cmake)
 
+include_directories(${PG_INCLUDE_DIR}) # for libpq
+
 GitAbbRev_Get(PLCONTAINER_GIT_VERSION)
 configure_file(
     "${PROJECT_SOURCE_DIR}/src/config.h.in"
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Gpdb.cmake)
 include(${CMAKE_ROOT}/Modules/ExternalProject.cmake)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/Git.cmake)
 
+include_directories(${PG_INCLUDE_DIR}) # for libpq
```
beeender, line 9: why this is needed? I think https://github.com/greenplum-db/plcontainer/pull/641/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR90 is enough?

concourse/scripts/entry.sh
```diff
@@ -167,19 +167,27 @@ function install_extra_build_dependencies() {
...
         yum install -y docker-ce docker-ce-cli postgresql-devel
         ;;
-    rhel8) ;;
-
-    ubuntu*) ;;
+    ubuntu*)
+        # ubuntu also need to install latest docker
+        apt update
+        apt install -y apt-transport-https ca-certificates curl software-properties-common
+        apt install -y libpq-dev # for libpq
```
beeender, line 174: hmmm. libpq-dev and postgresql-devel in line 168 should not be needed. Why they are there? all the header files should come from gpdb installation.

yihong0618: fixed

CMakeLists.txt
yihong0618: this is needed like `diskquota`

concourse/scripts/entry.sh
```diff
@@ -167,19 +167,26 @@ function install_extra_build_dependencies() {
         yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
         yum install -y docker-ce docker-ce-cli postgresql-devel
```
beeender, line 168: postgresql-devel should be removed as well.

beeender (approved):

PR #645, opened 2022-07-18
strict permission check when user enables network or GPU.

Sasasu:
if GPU or network not enabled, empty roles list means any user can use
    this runtime.

    if GPU or network enable, empty roles list means no one can use this
    runtime.


src/plc_configuration.c
```diff
@@ -282,6 +282,9 @@ static void parse_runtime_configuration(xmlNode *node) {
 									"changing <enable_network> to \"yes\" automaticly");
 							conf_entry->enableNetwork = true;
 						}
+
+						// network will a new attack surface. enable strict permission check
+						conf_entry->useUserControl = true;
```
beeender, line 287: This is not actually a good place for setting the config. The `enableNetwork` could be `false` at line 269.

I propose to have a unified place to set this at the end of this function/loop like:

```
if (conf_entry->enableNetwork || not_empty(roles) || enable_gpu) {
    conf_entry-> useUserControl = true;
}
```

src/plc_configuration.c
```diff
@@ -619,7 +629,10 @@ static void print_runtime_configurations() {
 				}
 			}
 		}
-		if (conf_entry->useUserControl && conf_entry->roles != NULL) {
+		if (conf_entry->useUserControl) {
+			plc_elog(INFO, "    useUserControl  = 'true'");
```
beeender, line 633: very minor. there are two spaces between `useUserControl` and `=`. So is line 636

src/plc_configuration.c
```diff
@@ -848,15 +861,24 @@ char *get_sharing_options(runtimeConfEntry *conf, int container_slot, bool *has_
...
 
 bool plc_check_user_privilege(char *roles){
-
 	List *elemlist;
 	ListCell *l;
 	Oid currentUserOid;
 
+	if (roles == NULL) {
+		ereport(WARNING,
+				(errmsg("plcontainer: enable network access or physical device access need set a explicit user permission for the runtime"),
```
beeender, line 870: To:

To access network or physical device, user permission needs to be granted explicitly in the runtime config.

tests/data/test_assign_gpu.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/env python2
+#!/usr/bin/env python2
```
beeender, line 1: hmmm, `/bin/env` has problems on some distros? which distro?

Sasasu: ubuntu18.04

src/plc_configuration.c
Sasasu: I want the check-and-fix section to be closer to the read-and-set section, so it's easier to distinguish dependencies. It is also easier when we split this large function.

fix `useUserControl = true when enableNetwork = 'no'`

beeender (approved):

src/plc_configuration.c
```diff
@@ -848,15 +863,24 @@ char *get_sharing_options(runtimeConfEntry *conf, int container_slot, bool *has_
...
 
 bool plc_check_user_privilege(char *roles){
-
 	List *elemlist;
 	ListCell *l;
 	Oid currentUserOid;
 
+	if (roles == NULL) {
+		ereport(WARNING,
+				(errmsg("plcontainer: To access network or physical device, user permission needs to be granted explicitly in the runtime config."),
```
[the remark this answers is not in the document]
Sasasu: I prefer not to use comment. need refactor the code struct.

split the very big config read logic `parse_runtime_configuration` and move the check-and-fix section to be closer to the read-and-set section.

need a refactor, not in this PR.

PR #649, opened 2022-08-05
fix: btrfs storage leak

yihong0618:


beeender:
> Otherwise, files created inside the container will be leaked to the concourse worker host.

Add some more comments

beeender (approved):

Sasasu (approved):

PR #650, opened 2022-08-09
[CI] Fetch secrets from vault.

zhrt123:
## Description

This patch teaches concourse to fetch secrets from vault.


beeender:
The diff doesn't look correct and the conflict need to be resolved. Did you create the PR based on a wrong branch?

PR #656, opened 2022-09-27
fix: bring python2.7 client back

yihong0618:
this `PR` bring python2.7 client back.

1. add cmake command `make py2client`
2. add dockerfile for python2.7 ubuntu18.04 build
3. add the same logic for python2.7 as python3.7
4. bring back python2.7 test
5. add version test(python2 or python3)

---

test file name rules:
python means --> python3
python2 means --> python2
e.g.
```
test_python_error means test python3 error in psql
test_pyhon2_error means test python2 error in psql
```


CMakeLists.txt
```diff
@@ -192,6 +228,22 @@ add_custom_target(pyclient
...
+    PYTHONPATH=
+    docker run
+    --entrypoint=
+    -v ${CMAKE_CURRENT_BINARY_DIR}/py2client:/build
+    -v ${CMAKE_CURRENT_SOURCE_DIR}:/src
+    -w /build
+    -u ${my_UID}
+    -it ${PY2CLIENT_IMAGE}
+    sh -c 'cmake /src/src/pyclient && cmake --build .')
+add_dependencies(py2client py2client_image)
```
beeender, line 245: Is there any special reason why the `py2client` is not added to the `clients` target at line 262?

tests/regress/pl_schedule_py2
```diff
@@ -0,0 +1,24 @@
+# setup - need to be first
```
beeender, line 1: shall we remove `tests/pl_schedule_py`? 

I am a bit confused when there is no version number following py/pythong. I think agree if there is no version following, then it is for python3? but the `ude_python.out` seems to be for python2?

yihong0618: yes we can.
If there's no version number it both works for python2 and python3

CMakeLists.txt
yihong0618: no special reason will add it.

beeender (approved):

PR #660, opened 2022-10-21
fix: release pipeline is publishing wrong intermediates

yihong0618:
**Same problem as https://github.com/greenplum-db/diskquota/pull/221**

Release pipeline is publishing wrong intermediates

"get" without "passed" will retrieve the latest content from a resource
which is not expected release. Especially the release and dev pipelines
share the intermediates bucket.

- Add passed the "get" step in the "exit_release" job, to make sure the
resource version generated from previous job will be consumed in this
job.


beeender (approved):
:+1:

PR #670, opened 2023-01-31
ci: switch gp binary to release-candidate for release build

Sasasu:
the binary in `server/published` has --enable-debug-extensions. but we do not want this. switch to `server/release-candidates.

but a binary with version 7.99.99 in release candidates was used for releng test. so some long regex string exists in this PR to filter out this version.

and skip faultinject test when for release build, faultinject is not enabled in gpdb release version.

### DIFF
--- a/concourse/pipeline/res_def.yml
+++ b/concourse/pipeline/res_def.yml
@@ -176,52 +176,40 @@ resources:
     bucket: pivotal-gpdb-concourse-resources-prod
     json_key: ((concourse-gcs-resources-service-account-key))
     regexp: server/published/gpdb6/server-rc-(.*)-ubuntu18.04_x86_64.debug.tar.gz
-# gpdb7
 - name: bin_gpdb7_rocky8_debug
   type: gcs
   source:
     bucket: pivotal-gpdb-concourse-resources-prod
     json_key: ((concourse-gcs-resources-service-account-key))
     regexp: server/published/main/server-rc-(.*)-rocky8_x86_64.debug.tar.gz
-- name: bin_gpdb7_rhel8_debug
-  type: gcs
-  source:
-    bucket: pivotal-gpdb-concourse-resources-prod
-    json_key: ((concourse-gcs-resources-service-account-key))
-    regexp: server/published/main/server-rc-(.*)-rhel8_x86_64.debug.tar.gz
+
 # Latest release candidates, no fault-injector, no assertion:
 # --disable-debug-extensions --disable-tap-tests --enable-ic-proxy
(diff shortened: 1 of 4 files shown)


concourse/pipeline/res_def.yml
```diff
@@ -176,52 +176,40 @@ resources:
...
 # Latest release candidates, no fault-injector, no assertion:
 # --disable-debug-extensions --disable-tap-tests --enable-ic-proxy
+# for the version string regexp: match every version but not {6,7}.99. .99 used for test
 - name: bin_gpdb6_centos7
   type: gcs
   source:
     bucket: pivotal-gpdb-concourse-resources-prod
     json_key: ((concourse-gcs-resources-service-account-key))
-    regexp: server/published/gpdb6/server-rc-(.*)-rhel7_x86_64.tar.gz
+    regexp: server/release-candidates/gpdb6/greenplum-db-server-6\.([0-9]|([1-8][0-9])|(9[0-8]))\.\d{1,2}-.*-centos6.tar.gz
```
higuoxing, line 194: The original server binary is centos7, but the current one seems centos6?

tests/CMakeLists.txt
```diff
@@ -1,5 +1,12 @@
 include(${CMAKE_SOURCE_DIR}/cmake/Regress.cmake)
 
+list(APPEND REGRESS_OPTS "--dbname=${TEST_DB_NAME}")
+if("$ENV{ARTIFACT_TYPE}" STREQUAL "Release")
```
higuoxing, line 4: It might be good not to rely on environment variables in our CI. I think we can add an additional build option to control whether build with fault injector and run isolation2 tests? e.g.,

```
cmake -DENABLE_ISOLATION2_TESTS=ON ## If this option is enabled, we'll build with fault injector and run isolation2 tests
```

We can revise this later. Just some thoughts of my own.

Sasasu: https://github.com/greenplum-db/gpdb/blob/a42310675a18df55e6ca31fc8a6344007fff456e/src/Makefile.global.in#L523-L539

this is simulating gpdb behavior. if faultinjector is not enabled, skip all related to faultinjector.
if `ARTIFACT_TYPE` is empty, will run the test with faultinjector, as same as before.

higuoxing: Yes, I mean the `ARTIFACT_TYPE` is CI-specific, I would like to have a cmake option to control it. Or, at least have a meaningful name for `ARTIFACT_TYPE`.

higuoxing (approved):

higuoxing:
LGTM! thanks!

PR #671, opened 2023-02-01
management: select python3 if no python or python2 available

Sasasu:
how does it work?

`management/bin/plcontainer` is both a shell script and a python script.

first this script will use the shell part to find python3 or python2. then use python and exec syscall to boot self.

the header `''':'` and `'''#'''` is some empty string and comment syntax in bash and python

### DIFF
--- a/management/bin/plcontainer
+++ b/management/bin/plcontainer
@@ -1,4 +1,20 @@
-#!/usr/bin/env python
+#!/bin/sh
+
+''':'
+PYTHON=$(head -n 1 "$GPHOME/lib/python/gppylib/gpcatalog.py" 2> /dev/null | cut -d ' ' -f 2)
+
+if [ -z $PYTHON ]; then
+    PYTHON=$(which python3 2> /dev/null || which python 2> /dev/null || which python2 2> /dev/null)
+fi
+
+if [ -z $PYTHON ]; then
+    echo "no python3 or python2 detected"
+    exit 1
+fi
+
+exec $PYTHON $0 "$@"
+exit $?
+''' #'''
 
 #------------------------------------------------------------------------------
 #
--- a/management/bin/plcontainer
+++ b/management/bin/plcontainer
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 ''':'
-PYTHON=$(head -n 1 "$GPHOME/lib/python/gppylib/gpcatalog.py" 2> /dev/null | cut -d ' ' -f 2)
+PYTHON=$(head -n 1 "$GPHOME/lib/python/gppylib/gplog.py" 2> /dev/null | cut -d ' ' -f 2)
 
 if [ -z $PYTHON ]; then
     PYTHON=$(which python3 2> /dev/null || which python 2> /dev/null || which python2 2> /dev/null)


management/bin/plcontainer
```diff
@@ -1,4 +1,16 @@
-#!/usr/bin/env python
+#!/bin/sh
+
+''':'
+PYTHON=$(which python3 2> /dev/null || which python 2> /dev/null || which python 2> /dev/null)
```
yihong0618, line 4: which python 2> /dev/null || which python 2> /dev/null same here?

yihong0618 (approved):
so neat!

management/bin/plcontainer
```diff
@@ -1,4 +1,20 @@
-#!/usr/bin/env python
+#!/bin/sh
+
+''':'
+PYTHON=$(head -n 1 "$GPHOME/lib/python/gppylib/gpcatalog.py" 2> /dev/null | cut -d ' ' -f 2)
+
+if [ -z $PYTHON ]; then
+    PYTHON=$(which python3 2> /dev/null || which python 2> /dev/null || which python2 2> /dev/null)
```
beeender, line 7: Since plcontainer relies on the gppylib, just throw errors if the python executable cannot be got from the gppylib source.

The guess here mostly won't work I think.

management/bin/plcontainer
```diff
@@ -1,4 +1,20 @@
-#!/usr/bin/env python
+#!/bin/sh
+
+''':'
+PYTHON=$(head -n 1 "$GPHOME/lib/python/gppylib/gpcatalog.py" 2> /dev/null | cut -d ' ' -f 2)
```
beeender, line 4: Choose a file we import here? Maybe `gplog.py`. If the `gplog.py` doesn't exist, this file won't work anyway.

management/bin/plcontainer
```diff
@@ -1,4 +1,20 @@
...
+    PYTHON=$(which python3 2> /dev/null || which python 2> /dev/null || which python2 2> /dev/null)
+fi
+
+if [ -z $PYTHON ]; then
+    echo "no python3 or python2 detected"
+    exit 1
+fi
+
+exec $PYTHON $0 "$@"
+exit $?
```
beeender, line 16: Is this needed? exec will execute the python in the current process and exit the current process. I don't think `exit $?` will get a chance to run.

management/bin/plcontainer
Sasasu: the guess will select a useable python executable and emit the error message from the python side.
```python
except ImportError as e:
  sys.exit('ERROR: Cannot import Greenplum modules.  Please check that you have sourced greenplum_path.sh.  Detail: ' + str(e))
```

management/bin/plcontainer
beeender: > exit 1 if $PYTHON is not an executable file

hmmm, interesting, bash and zsh have different behaviors.

Bash will report

```
[cc@cc-desktop ~]$ exec asdfasdf
bash: exec: asdfasdf: not found
```


well, zsh will just terminate the current shell.

`exit $?` here is good for bash. :+1:

beeender (approved):

PR #679, opened 2023-07-17
feat: change base dockerfile from python3.7 -> python3.9 

yihong0618:
## Description

for now, python3.7 is end of life, we change dockerfile and data-science-bundle from python3.7 to python3.9, also change ubuntu1804 -> ubuntu2204


src/pyclient/Makefile
```diff
@@ -10,7 +10,7 @@
 PLCONTAINER_DIR = ..
 
 ifeq ($(PYTHON_VERSION),3)
-  PYTHON = python3.7
+  PYTHON = python3.9
```
beeender, line 13: no need to change this. we use cmake to build

yihong0618 (commented):
LGTM

PR #665, opened 2023-01-06
Make PL/Container compatible with Greenplum 7.

higuoxing:
break change due to PostgreSQL break change:

- `select f1(), f2()` will not start a join execution.
-  when `elog(ERROR)` happens, all log messages will be truncated, the log stream is not realtime.
- not support memory limit anymore. GP7 does not support resource group

the plcontainer source code is GP6/GP7 compatible.
all released client binary will support both GP6 and GP7

PR #691, opened 2023-09-21
management: fix runtime-edit crash in py3

Sasasu:
fllow up https://github.com/greenplum-db/plcontainer/pull/665

```
Traceback (most recent call last):
  File "/home/sa/GPDB/plcontainer/./management/bin/plcontainer", line 974, in main
    changed = edit_file(tmp_file, options.editor)
  File "/home/sa/GPDB/plcontainer/./management/bin/plcontainer", line 647, in edit_file
    h1 = get_file_md5(filename)
  File "/home/sa/GPDB/plcontainer/./management/bin/plcontainer", line 486, in get_file_md5
    h.update(l)
TypeError: Strings must be encoded before hashing
```

### DIFF
--- a/management/bin/plcontainer
+++ b/management/bin/plcontainer
@@ -483,7 +483,7 @@ def get_file_md5(file):
     h = hashlib.md5()
     f = open(file, 'r')
     for l in f:
-        h.update(l)
+        h.update(l.encode('utf-8'))
     f.close()
     return h.digest()
 
--- a/management/bin/plcontainer
+++ b/management/bin/plcontainer
@@ -480,12 +480,8 @@ def get_file_md5(file):
     of bytes and cannot be printed directly to the screen or log file
     without converting to hex.
     '''
-    h = hashlib.md5()
-    f = open(file, 'r')
-    for l in f:
-        h.update(l.encode('utf-8'))
-    f.close()
-    return h.digest()
+    with open(file, 'rb') as f:
+        return hashlib.md5(f.read()).digest()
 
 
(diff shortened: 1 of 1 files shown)


management/bin/plcontainer
```diff
@@ -483,7 +483,7 @@ def get_file_md5(file):
     h = hashlib.md5()
     f = open(file, 'r')
     for l in f:
-        h.update(l)
+        h.update(l.encode('utf-8'))
```
higuoxing, line 486: Seems that we can simplify it to:

```
def get_file_md5(file):
    with open(file, 'rb') as f:
        return hashlib.md5(f.read()).digest()
```

> hash.update(data)
> Update the hash object with the [bytes-like object](https://docs.python.org/3/glossary.html#term-bytes-like-object). Repeated calls are equivalent to a single call with the concatenation of all the arguments: m.update(a); m.update(b) is equivalent to m.update(a+b).

Ref: https://docs.python.org/3/library/hashlib.html#hashlib.hash.update

Sasasu: from py2.7: `Update the hash object with the string arg.`
I think this does not support python2

Sasasu: 401ecfe5

higuoxing (approved):

yihong0618:
nice rewitre, and I think all the open close things can replace to `with open() as f` in this file

Sasasu:
test in both gp6 with py2 and gp7 with py3

PR #693, opened 2023-09-22
Fix the python3 image name

beeender:
which should be aligned with pivnet.


yihong0618 (approved):
so for now, python3 is python3
python2 is python2 for the image name ?

Sasasu (approved):

Sasasu:
image

in `runtime add -l` python is python2, are you sure we need a rename?

PR #698, opened 2023-10-23
Use ci common entry.sh

RMTT:
Use ci common entry.sh to keep consistency.


cmake/Distro.cmake
```diff
@@ -16,12 +16,17 @@ if(NOT DISTRO_NAME)
         string(REGEX MATCH "Red Hat Enterprise Linux release 8.*" matched_rhel8 "${rh_release}")
         string(REGEX MATCH "CentOS Linux release 8.*" matched_centos8 "${rh_release}")
         string(REGEX MATCH "Rocky Linux release 8.*" matched_rocky8 "${rh_release}")
+        string(REGEX MATCH "Red Hat Enterprise Linux release 9.*" matched_rhel9 "${rh_release}")
```
beeender, line 19: I think those have been merged in #697 ?

concourse/scripts/docker-lib.sh
```diff
@@ -84,7 +84,6 @@ start_docker() {
   fi
 
   export server_args LOG_FILE DOCKER_DATA_ROOT
-  trap stop_docker EXIT
```
beeender, line 87: This ensures the inner docker won't be leaked when the process exists.

Any reason why this is removed?

concourse/scripts/pre_test.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -exo
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+source "$SCRIPT_DIR/pre_build.sh"
+
+# print the test diff to stdout in our CI
+export SHOW_REGRESS_DIFF=1
```
beeender, line 10: This should be moved to `test.sh`. Setting this in the `root` session doesn't do anything, since the tests will be run in the `gpadmin` session.

cmake/Distro.cmake
RMTT: yeah, i forgot clean my local branch.

concourse/scripts/docker-lib.sh
RMTT: because this script run in `pre-test/build.sh` now, so when run `test/buid.sh`, the docker has been stopped. I think i can put `start_docker` in `test/build.sh`.

beeender: >  the docker has been stopped. I think i can put start_docker in test/build.sh

This should work. Please verify if the container has been cleaned when the `test/build.sh` exits.

beeender (approved):
Merge it when CI is happy.

PR #707, opened 2023-12-06
fix cmake error

RMTT:
CMake 3.27.9 in [pypi](https://pypi.org/project/cmake/#history) broke python2 support, but 3.27.7 is ok.

### DIFF
--- a/dockerfiles/Dockerfile.python27.ubuntu
+++ b/dockerfiles/Dockerfile.python27.ubuntu
@@ -3,9 +3,9 @@ FROM ubuntu:18.04
 EXPOSE 8080
 RUN mkdir -p /clientdir
 RUN apt update && apt install -y python2.7 python-pip python-dev gcc g++ make
-# for easily install the latest cmake
-RUN pip install -U pip
-RUN pip install cmake==3.27.7
+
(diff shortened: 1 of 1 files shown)


dockerfiles/Dockerfile.python27.ubuntu
```diff
@@ -5,7 +5,7 @@ RUN mkdir -p /clientdir
 RUN apt update && apt install -y python2.7 python-pip python-dev gcc g++ make
 # for easily install the latest cmake
 RUN pip install -U pip
-RUN pip install cmake
+RUN pip install cmake==3.27.7
```
beeender, line 8:
```
ADD https://github.com/Kitware/CMake/releases/download/v3.26.5/cmake-3.26.5-linux-x86_64.sh /cmake.sh
RUN sh /cmake.sh --prefix=/usr/local/ --exclude-subdir --skip-license && rm /cmake.sh 
``` 

Do this, then we don't have to worry about it anymore.

https://github.com/pivotal/gp-data-science-bundle/blob/bbd7d262c0842c10636f577c26780e44d39c4ea1/dockerfiles/r/Dockerfile.r.ubuntu1804_b#L11

beeender: pip is not always reliable

beeender (approved):

PR #709, opened 2023-12-11
management: fix help message. copy greenplum_path.sh to remove. export 8080/tcp

Sasasu:
ip list only support `,`

### DIFF
--- a/management/bin/plcontainer
+++ b/management/bin/plcontainer
@@ -87,11 +87,11 @@ def parseargs():
     image_add_options.add_argument("-f", "--file", dest="file", help="image location (e.g. /tmp/centos.7.2.tar.gz)")
     image_add_parser.add_argument("-ulc", "--use_local_copy", action="store_true", dest="local", help="If set, use the local image file (default is no)")
     image_add_options.add_argument("-u", "--url", dest="url", help="image url (e.g. 192.168.0.1:5000/images/centos:7.2)")
-    image_add_parser.add_argument("--hosts", dest="hosts", default='', help="also load image for those IP in this host IP list (e.g. '128.66.0.1 128.66.0.2,128.66.0.3') or a gpscp hostilfe")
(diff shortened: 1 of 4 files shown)


radarwave (approved):

management/bin/plcontainer
```diff
@@ -330,6 +346,7 @@ def setup_remote(pool, dest_dir, hosts):
     try:
         logger.info('Copying clientdir to %s on all hosts' % (dest_dir,))
         distribute_clientdir(pool, dest_dir, hosts)
+        distribute_greenplum_path(pool, dest_dir, hosts)
```
RMTT, line 349: Why distribute `greenplum_path.sh` to remote host?

Sasasu: `pool.addCommand(cmd)` will execute this in side `ssh <ip> ". /usr/local/greenplum-db-7.0.0/greenplum_path.sh; <the command>"`
so the `greenplum_path.sh` is needed on the remote side. but we don't use it

RMTT (approved):

src/plc_configuration.c
```diff
@@ -614,8 +614,7 @@ static void parse_runtime_configuration(xmlNode *node) {
 					}
 
 					if (conf_entry->enableNetwork == false && conf_entry->backend->tag == PLC_BACKEND_REMOTE_DOCKER) {
-						plc_elog(WARNING, "runtime \"%s\": <enable_network> must be true when using remote_docker as backend."
-								" turnning on network automaticly", conf_entry->runtimeid);
+						plc_elog(ERROR, "runtime \"%s\": <enable_network> must be true when using remote_docker as backend." , conf_entry->runtimeid);
 						conf_entry->enableNetwork = true;
```
RMTT, line 618: Why remove the warning log but keep the `conf_entry->enableNetwork = true;`

